PR #3, opened 2020-02-14
Add Initial BFT Skeleton

talekhinezh:
* Add initial dumb implementations of Pacemaker, MemPool, Network, and EventCoordinator


Sajjon (commented):
Nice! Left some comments/questions

flotothemoon (approved):
Looks good enough for now, fix/consider my criticism at your own discretion

clstrfsck (commented):
Some comments.

Would also like to see more comments on public methods that we think will stay or thereabouts.

PR #11, opened 2020-02-20
Feature/rx java

talekhinezh:


flotothemoon (approved):
Generally fine with this, some comments on method naming but not critical.

clstrfsck (commented):
Do we need to consider use of Flowable vs Observable for things requiring backpressure?

PR #12, opened 2020-02-21
Feature/pacemaker tests

talekhinezh:


clstrfsck (commented):
A bit concerned that these unit tests actually have to wait for timeouts to happen to work correctly.  I think we should be making some attempt now to resolve the issue with using a singleThreadedExecutor.

Also concerned that we are creating a bunch of classes that have a start() method, which would suggest that they probably should have a stop() method too, but we are not implementing that.  This class needs a stop() method, otherwise there is no way to stop the thread it creates.  This is an issue for both runtime and unit tests.

I really think we should consider some kind of Round POJO now.  Code is getting hard to read.

clstrfsck (commented):
Some minor tidy-ups suggested, but nothing that affects behaviour.

PR #19, opened 2020-02-26
A QuorumCertificate now have a signatures variable, and a means to ag…

Sajjon:
…gregate matching votes and their signatures.


Sajjon:
This is a non-working **draft** PR.

@flotothemoon Please help me work out this mess 😅 Where should the aggegation of signatures take place?

PR #22, opened 2020-02-28
Feature/basic safety rules

flotothemoon:
Implements the basic safety rules of HotStuff as an initial step of the Cerberus implementation.

PR #24, opened 2020-03-03
Improvement/rpn1 48 equalsverifier

Sajjon:
Branched of #22, do not merge before that branch has been merged.

**Using #22 as target branch for comparison purposes.**

Draft PR of [RPNV1-48](https://radixdlt.atlassian.net/browse/RPNV1-48)


Sajjon:
Not my Code Smells, inherited from #22 .

radixdlt/src/main/java/com/radixdlt/store/LedgerEntry.java
```diff
@@ -68,12 +76,13 @@ public boolean equals(Object o) {
...
 		LedgerEntry radixLedgerEntry = (LedgerEntry) o;
-		return aid.equals(radixLedgerEntry.aid);
+		return Objects.equals(aid, radixLedgerEntry.aid)
+				&& Arrays.equals(content, radixLedgerEntry.content);
 	}
 
 	@Override
 	public int hashCode() {
-		return aid.hashCode();
+		return Objects.hash(aid, Arrays.hashCode(content));
```
clstrfsck: This isn't as elegant as it could be.  What happens here is that `Arrays.hashCode(byte[])` calculates a primitive `int` hash, this is then boxed to an `Integer` and rehashed and combined with the hash of aid.

Would be better as something like `return Objects.hashCode(aid) ^ Arrays.hashCode(content)`, or more standardly `return Objects.hashCode(aid) * 31 + Arrays.hashCode(content)`.

In any case, it might be better to consider excluding the content from the equals and hashCode methods, as the AID should be (barring hash collisions) a representation of the content.

radixdlt/src/main/java/org/radix/utils/Range.java
```diff
@@ -54,14 +54,18 @@ protected void setHigh(T high)
 	}
 
 	@Override
-	public boolean equals(Object other)
+	public final boolean equals(Object other)
 	{
 		if (other == this)
 			return true;
 
 		if (other == null)
```
clstrfsck, line 62: Not your code, but `null` is never an `instanceof` anything, so this check is redundant.

radixdlt/src/main/java/com/radixdlt/store/LedgerEntry.java
```diff
@@ -68,12 +76,13 @@ public boolean equals(Object o) {
 			return false;
 		}
 		LedgerEntry radixLedgerEntry = (LedgerEntry) o;
-		return aid.equals(radixLedgerEntry.aid);
+		return Objects.equals(aid, radixLedgerEntry.aid)
```
clstrfsck: See comment about content in hashCode.

clstrfsck (commented):
Looks good.  Awesome code coverage :)  Minor comment on hashing/equals for LedgerEntry.

radixdlt/src/main/java/com/radixdlt/store/LedgerEntry.java
Sajjon: removed `content`, and ignoring it in EqualsVerifier test

radixdlt/src/main/java/com/radixdlt/store/LedgerEntry.java
Sajjon: removed `content`, and ignoring it in EqualsVerifier test

radixdlt/src/main/java/org/radix/utils/Range.java
Sajjon: removed `if (other == null); return false`

Sajjon: @clstrfsck Spoke to soon, actually this seems needed, at least I did not get it to work by changing this:

```java
		if (other == null)
			return false;

		// `FooRange` should not automatically equal `(Range) BarRange`.
		if (!this.getClass().equals(other.getClass()))
			return false;
```

into just this:

```java
		// `FooRange` should not automatically equal `(Range) BarRange`.
		if (!Objects.equals(this.getClass(), other.getClass()))
			return false;
```

Getting error:
```
EqualsVerifier found a problem in class ShardRange.
-> Non-nullity: NullPointerException thrown
```

And not using `getClass` results in EqualsVerifier error relating to inheritance:

```
EqualsVerifier found a problem in class ShardRange.
-> Redefined superclass:
  nl.jqno.equalsverifier.internal.reflection.Number$$DynamicSubclass$1706292388@2ff99a17 -> nl.jqno.equalsverifier.internal.reflection.Number$$DynamicSubclass$1706292388@2ff99a17
should not equal superclass instance
  nl.jqno.equalsverifier.internal.reflection.Number$$DynamicSubclass$1706292388@2ff99a17 -> nl.jqno.equalsverifier.internal.reflection.Number$$DynamicSubclass$1706292388@2ff99a17
but it does.
```

Basically we don't want `FooRange extends Range` and `BarRange extends Range` to equal one and other.

PR #26, opened 2020-03-04
Initialise genesis QC if no highest QC exists

flotothemoon:
Instead of using a null QC use a proper genesis QC (with all zeros as rounds & id) as the first QC in our tree.


Sajjon (approved):
LGTM!

PR #29, opened 2020-03-13
Feature/update re api

talekhinezh:


clstrfsck (changes requested):
Looks mostly good, but some checkstyle issues causing travis to break:
https://travis-ci.com/github/radixdlt/radixdlt-core/jobs/297871123#L780

PR #31, opened 2020-03-17
Feature/vertex store update

talekhinezh:


radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -83,13 +77,11 @@ private void processNewRound(Round round) {
...
 
-		List<Atom> atoms = mempool.getAtoms(1, Sets.newHashSet());
-		if (!atoms.isEmpty()) {
-			QuorumCertificate highestQC = vertexStore.getHighestQC()
-				.orElseGet(this::makeGenesisQC);
+		Vertex proposal = proposalGenerator.generateProposal(this.pacemaker.getCurrentRound());
 
-			log.info("Starting round " + round + " with proposal " + atoms.get(0));
-			networkSender.broadcastProposal(new Vertex(highestQC, this.pacemaker.getCurrentRound(), atoms.get(0)));
+		// TODO: Handle empty proposals
```
clstrfsck, line 82: Wouldn't it be best to handle this now, rather than just letting the pipeline stall?

radixdlt/src/main/java/com/radixdlt/consensus/Vertex.java
```diff
@@ -57,12 +57,40 @@
...
 		this.atom = atom;
 	}
 
+	public static Vertex createGenesis(Atom atom) {
+		return new Vertex(null, Round.of(0), atom);
+	}
+
+	public static Vertex createVertex(QuorumCertificate qc, Round round, Atom atom) {
+		Objects.requireNonNull(qc);
+		Objects.requireNonNull(round);
```
clstrfsck: This isn't really required, because the following test will throw an NPE immediately in any case.

radixdlt/src/main/java/com/radixdlt/consensus/Vertex.java
```diff
@@ -87,7 +111,7 @@ private Long getSerializerRound() {
 
 	@JsonProperty("round")
 	private void setSerializerRound(Long number) {
-		this.round = number == null ? null : Round.of(number.longValue());
+		this.round = number == null ? null : Round.of(number);
```
clstrfsck, line 127: Why remove the explicit unboxing here?

radixdlt/src/main/java/com/radixdlt/consensus/VertexInsertionException.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * An exception indicating a failure in inserting a vertex into a VertexStore
+ */
+class VertexInsertionException extends Exception {
+	VertexInsertionException(String message) {
+		super(message);
+	}
+
+	VertexInsertionException(Exception cause) {
```
clstrfsck: This constructor should require a context message as well.  If just the cause is included, it can be hard to determine what went wrong without a context message.

radixdlt/src/main/java/com/radixdlt/consensus/Vertex.java
```diff
@@ -57,12 +57,40 @@
...
+		if (round.number() == 0) {
+			throw new IllegalArgumentException("Only genesis can have round 0.");
+		}
+
+		return new Vertex(qc, round, atom);
+	}
+
+	// TODO: Replace with a better type over EUID wrapping hashCode()
+	public EUID getId() {
+		return new EUID(this.hashCode());
```
clstrfsck: Can we fix this?  Using a tiny 32 bit hash of dubious merit seems almost certain to bite us hard at some point in the near future.

Given that this class is now serializable, it should not be too hard to calculate a proper hash / EUID.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -43,12 +59,26 @@ public void syncToQC(QuorumCertificate qc) {
...
+		}
+
 		this.syncToQC(vertex.getQC());
-		vertices.add(vertex);
+
+		if (vertex.getAtom() != null) {
+			try {
+				this.engine.store(vertex.getAtom());
+			} catch (RadixEngineException e) {
+				throw new VertexInsertionException(e);
```
clstrfsck: Needs context message.  Should be something like:
```
throw new VertexInsertionException("While storing atom", e);
```

radixdlt/src/main/java/com/radixdlt/consensus/liveness/ProposalGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Objects;
+
+/**
+ * Logic for generating new proposals
+ */
+public final class ProposalGenerator {
+	private final Mempool mempool;
+	private final VertexStore vertexStore;
+
+	@com.google.inject.Inject
```
clstrfsck: Would be better to use @Inject rather than fully qualified name.

Would also be better to use javax.inject version, as this is not specific to guice.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/ProposalGenerator.java
```diff
@@ -0,0 +1,37 @@
...
+	) {
+		this.vertexStore = Objects.requireNonNull(vertexStore);
+		this.mempool = Objects.requireNonNull(mempool);
+	}
+
+	// TODO: check that next proposal works with current vertexStore state
+	public Vertex generateProposal(Round round) {
+		QuorumCertificate highestQC = vertexStore.getHighestQC();
+
+		List<Atom> atoms = mempool.getAtoms(1, Sets.newHashSet());
```
clstrfsck: I feel like I have complained about this a lot, but never achieved a satisfactory resolution, nor received a satisfactory response.

The second argument to `mempool.getAtoms(...)` *cannot be a fresh set every time*, or bad things will happen.  In particular if it does not include all of the atom IDs for atoms in the pipeline, then duplicates will be proposed.

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
```diff
@@ -50,4 +65,23 @@ protected void configure() {
...
+		Universe universe,
+		RadixEngine radixEngine,
+		@Named("self") EUID self
+	) throws RadixEngineException {
+		if (universe.getGenesis().size() != 1) {
+			throw new IllegalStateException("Can only support one genesis atom.");
+		}
+
+		final Vertex genesisVertex = Vertex.createGenesis(universe.getGenesis().get(0));
+		final VertexMetadata genesisMetadata = new VertexMetadata(Round.of(0), genesisVertex.getId(), Round.of(0), genesisVertex.getId());
```
clstrfsck: Would be better if `Round.of(0)` was replaced with something like `Round.GENESIS`, or `Round.genesis()`.

clstrfsck (changes requested):
Some checkstyle issues need addressing, see https://travis-ci.com/github/radixdlt/radixdlt-core/jobs/299019245#L986

Would really like to see us address the issue of using a tiny 32 bit hashCode() as an ID in Vertex.

Concerned that we are not properly maintaining our set of seen atom IDs in generateProposal(...).

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
talekhinezh: Problem with handling it now is that because it is single node+responsiveness, rounds will just continue to increase as fast as it can making it harder to debug, so avoiding this change until later.

radixdlt/src/main/java/com/radixdlt/consensus/Vertex.java
talekhinezh: IntelliJ suggestion as it is apparently unnecessary

radixdlt/src/main/java/com/radixdlt/consensus/Vertex.java
talekhinezh: Yes, we also perhaps need a discussion on whether 128 bits is enough vs 256 bits like bitcoin

radixdlt/src/main/java/com/radixdlt/consensus/liveness/ProposalGenerator.java
talekhinezh: Right, will add a fix in this PR.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/ProposalGenerator.java
```diff
@@ -0,0 +1,38 @@
...
+		VertexStore vertexStore,
+		Mempool mempool
+	) {
+		this.vertexStore = Objects.requireNonNull(vertexStore);
+		this.mempool = Objects.requireNonNull(mempool);
+	}
+
+	// TODO: check that next proposal works with current vertexStore state
+	public Vertex generateProposal(Round round) {
+		QuorumCertificate highestQC = vertexStore.getHighestQC();
```
flotothemoon: As discussed, I have some doubts that this is the best way to get the QC to extend. Fine for now, we can revisit when I integrate my branch.

radixdlt/src/main/java/com/radixdlt/consensus/Vertex.java
```diff
@@ -50,17 +53,55 @@
...
+		if (round.number() == 0) {
+			throw new IllegalArgumentException("Only genesis can have round 0.");
+		}
+
+		return new Vertex(qc, round, atom);
+	}
+
+	private Hash doGetHash() {
+		try {
+			return new Hash(Hash.hash256(Serialization.getDefault().toDson(this, Output.HASH)));
```
flotothemoon, line 89: Do we really want a static dependency on Serialization here?

flotothemoon (commented):
Some general comments, nothing crucial. Apart from what Martin already mentioned this is fine by me.

radixdlt/src/main/java/com/radixdlt/consensus/Vertex.java
talekhinezh: probably not, but using the pattern we've been using for all other classes is fine for now until we come up with better strategy.

radixdlt/src/main/java/com/radixdlt/consensus/MissingParentException.java
```diff
@@ -0,0 +1,14 @@
...
+import com.radixdlt.crypto.Hash;
+import java.util.Objects;
+
+/**
+ * Exception specifying that a vertex cannot be inserted because
+ * it's parent is missing from the current store.
+ */
+class MissingParentException extends VertexInsertionException {
+	MissingParentException(Hash parentId) {
+		super("Parent Vertex missing: " + Objects.requireNonNull(parentId));
```
clstrfsck, line 12: Really not sure the null check here is a good idea.  It seems a bit of an implementation smell to have construction of an exception, which will presumably be thrown shortly, potentially throw a NullPointerException.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -43,12 +63,48 @@ public void syncToQC(QuorumCertificate qc) {
...
+
+		committedVertices.put(vertexId, vertex);
+		return vertex;
+	}
+
+	public List<Vertex> getPathFromRoot(Hash vertexId) {
+		List<Vertex> path = new ArrayList<>();
+
+		Vertex vertex = vertices.get(vertexId);
+		while (!committedVertices.containsKey(vertex.getId())) {
```
clstrfsck: Should probably be `while (vertex != null && !committedVertices.containsKey(vertex.getId())) {` to avoid NPE on arbitrary hash being passed in as argument.

clstrfsck (approved):
I think the NPE hazard in `getPathFromRoot(...)` needs fixing, but good to go after that.

PR #32, opened 2020-03-18
Multi-node and signature support

flotothemoon:
Adds support for multiple nodes (quorums of size > 1) with signed and accumulated votes. Signatures are not verified yet and identity management (for address/pkey) should be made more robust. 

Also renames Round => View to align terminology with both our paper and the relevant literature. Naturally, this causes a lot of not-actually-changed changes, so feel free to ignore those.


radixdlt/src/main/java/com/radixdlt/CerberusModule.java
```diff
@@ -71,15 +78,15 @@ private PacemakerImpl pacemaker() {
 	private VertexStore getVertexStore(
 		Universe universe,
 		RadixEngine radixEngine,
-		@Named("self") EUID self
+		@Named("self") RadixAddress self
```
talekhinezh: What's the reasoning for using RadixAddress? RadixAddress is more of an application layer abstraction

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -58,7 +66,9 @@ public EventCoordinator(
 		Pacemaker pacemaker,
 		VertexStore vertexStore,
 		ProposerElection proposerElection,
-		@Named("self") EUID self
+		QuorumRequirements quorumRequirements,
+		@Named("self") RadixAddress selfAddress,
```
talekhinezh: Same comment as above

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -58,7 +66,9 @@ public EventCoordinator(
...
+		@Named("self") ECKeyPair selfKey
```
talekhinezh, line 67: Shouldn't this be in SafetyRules?

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -67,57 +77,80 @@ public EventCoordinator(
...
 
-		// accumulate votes into QCs
-		// TODO assumes a single node network for now
-		QuorumCertificate qc = new QuorumCertificate(vote, vote.getVertexMetadata());
-		this.safetyRules.process(qc);
-		this.vertexStore.syncToQC(qc);
-		this.pacemaker.processQC(qc.getRound())
-			.ifPresent(this::processNewRound);
+		// accumulate votes into QCs in store
+		Optional<QuorumCertificate> potentialQc = this.vertexStore.insertVote(vote, this.quorumRequirements);
```
talekhinezh: Don't think vote accumulation should be occurring in the vertexStore

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -135,21 +168,15 @@ public void processProposal(Vertex proposedVertex) {
...
+			final Vote vote = safetyRules.voteFor(proposedVertex);
 			networkSender.sendVote(vote);
-			voteResult.getCommittedVertexId()
-				.ifPresent(vertexId -> {
-					log.info("Committed vertex " + vertexId);
-
-					final Vertex vertex = vertexStore.commitVertex(vertexId);
-					final Atom committedAtom = vertex.getAtom();
-					mempool.removeCommittedAtom(committedAtom.getAID());
-				});
```
talekhinezh, line 150: Why is the logic above removed?

radixdlt/src/main/java/com/radixdlt/consensus/NewView.java
```diff
@@ -0,0 +1,108 @@
...
+		if (!(o instanceof NewView)) {
+			return false;
+		}
+		NewView newView = (NewView) o;
+		return Objects.equals(this.view, newView.view);
+	}
+
+	@Override
+	public int hashCode() {
+		return this.view.hashCode();
```
talekhinezh: Should include the other parameters otherwise they should be transient.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -58,11 +64,38 @@ public void syncToQC(QuorumCertificate qc) {
 			return;
 		}
 
-		if (highestQC == null || highestQC.getRound().compareTo(qc.getRound()) < 0) {
+		if (highestQC == null || highestQC.getView().compareTo(qc.getView()) < 0) {
 			highestQC = qc;
 		}
 	}
 
+	public Optional<QuorumCertificate> insertVote(Vote vote, QuorumRequirements quorumRequirements) {
```
talekhinezh: As mentioned, this is state which shouldn't be handled by VertexStore

radixdlt/src/main/java/com/radixdlt/consensus/Vote.java
```diff
@@ -38,36 +40,44 @@
 
 	@JsonProperty("author")
 	@DsonOutput(Output.ALL)
-	private final EUID author;
+	private final RadixAddress author;
```
talekhinezh: Don't think this should be RadixAddress which is an application level object

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -20,47 +20,91 @@
...
 	/**
-	 * Process a quorum certificate
+	 * Process a QC seen at a certain view
 	 * @param qc The quorum certificate
+	 * @return the now-committed aid, if any
 	 */
-	public void process(QuorumCertificate qc) {
-		if (qc.getParentRound().compareTo(this.state.lockedRound) > 0) {
-			this.state = this.state.withLockedRound(qc.getParentRound());
+	public Optional<Hash> process(QuorumCertificate qc) {
```
talekhinezh, line 69: Marked for Josh to re-look at logic

radixdlt/src/main/java/com/radixdlt/consensus/safety/VertexHasher.java
```diff
@@ -1,33 +1,25 @@
...
-	}
+import com.radixdlt.consensus.VertexMetadata;
+import com.radixdlt.crypto.Hash;
 
-	private static NewRoundMessage get() {
-		NewRound testRound = new NewRound(Round.of(1234567890L));
-		return new NewRoundMessage(1234, testRound);
-	}
+public interface VertexHasher {
+	Hash hash(VertexMetadata vertexMetadata);
```
talekhinezh: As noted there are now two hashes per Vertex, one a hash of Vertex and another a hash of VertexMetadata. Need to discuss this...

talekhinezh (changes requested):

radixdlt/src/main/java/com/radixdlt/consensus/safety/VertexHasher.java
flotothemoon: Yeah, that's definitely weird. I want to unify them as that's just confusing.

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
flotothemoon: I recall there being an implementation reason for this, but it no longer applies, so I've reverted this.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
flotothemoon: Agreed, where do you propose putting it? Should probably just go in its own `PendingVotes` class.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
flotothemoon: Wasn't removed, was just moved to another place - see lines 117:121

radixdlt/src/main/java/com/radixdlt/consensus/NewView.java
flotothemoon: Updated to include other params

radixdlt/src/main/java/com/radixdlt/consensus/Vote.java
flotothemoon: Yup, agreed, updated

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
flotothemoon: This is for signing new-views, so no. But I agree that it should be in a different place, this is what my comments about "unified identity management" were referring to - we're currently implementing the same hash/sign flow in both safety rules and here in EC.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
talekhinezh: Yes, that sounds about right.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
talekhinezh: Don't think this should be removed from here though. Processing a new proposal also means processing unseen QCs.

flotothemoon: You're totally right, that should be at the start though (as it's not dependent on all the other stuff)

radixdlt/src/main/java/com/radixdlt/consensus/safety/VertexHasher.java
flotothemoon: Resolved by removing VertexHasher and using Vertex.getId directly for now.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -57,70 +64,100 @@ public EventCoordinator(
...
+			log.warn(String.format("Ignoring confused vote %s for %s", vote.hashCode(), vote.getVertexMetadata().getView()));
 			return;
 		}
 
-		// accumulate votes into QCs
-		// TODO assumes a single node network for now
-		QuorumCertificate qc = new QuorumCertificate(vote, vote.getVertexMetadata());
-		this.safetyRules.process(qc);
+		// accumulate votes into QCs in store
+		Optional<QuorumCertificate> potentialQc = this.pendingVotes.insertVote(vote, this.quorumRequirements);
```
talekhinezh: quorumRequirements should be a depdendency of PendingVotes no?

flotothemoon: It seemed more flexible to have it as an argument, but you're right, making it a dependency for now is simpler.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -20,47 +20,81 @@
...
+	 */
+	public Optional<Hash> process(Vertex vertex) {
+		return process(vertex.getQC());
 	}
 
 	/**
-	 * Process a quorum certificate
+	 * Process a QC seen at a certain view
 	 * @param qc The quorum certificate
+	 * @return the now-committed aid, if any
```
talekhinezh: Not AID but the VertexId

flotothemoon: Good catch

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -20,47 +20,81 @@
...
+		// pre-commit phase on vertex's parent if there is a newer 1-chain
+		// keep highest 1-chain as the current "generic" QC
+		if (qc.getView().compareTo(this.state.getGenericView().orElse(View.of(0L))) > 0) {
+			this.state = this.state.withGenericQC(qc);
 		}
+
+		// commit phase on vertex's grandparent if there is a newer 2-chain
+		// keep the highest 2-chain as the locked QC
+		Vertex parent = vertexStore.getVertex(qc.getVertexMetadata().getId());
+		if (parent != null && parent.getQC() != null)  { // genesis QC may be null
```
talekhinezh: parent != null should be an IllegalStateException
parent.getQC(), genesis QC should never go through process() so this should be IllegalStateException as well

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -20,47 +20,81 @@
...
+			if (parent.getQC().getView().compareTo(this.state.getLockedView()) > 0) {
```
talekhinezh: I think you need to add the fact that the parentView must be qc.getView() - 1.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -20,47 +20,81 @@
...
+				this.state = this.state.withLockedView(parent.getQC().getView());
+			}
+
+			// decide phase on vertex's great-grandparent if there is a newer 3-chain
+			// return committed aid
+			Vertex grandparent = vertexStore.getVertex(parent.getQC().getVertexMetadata().getId());
+			if (grandparent != null && grandparent.getQC() != null) { // genesis QC may be null
+				boolean threeChain = qc.getVertexMetadata().getView() == parent.getView()
+					&& parent.getQC().getVertexMetadata().getView() == grandparent.getView();
```
talekhinezh: Also need consecutive views here

talekhinezh (changes requested):

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
flotothemoon: Implemented on latest

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
```diff
@@ -62,24 +66,24 @@ protected void configure() {
...
+		return new PacemakerImpl(quorumRequirements, Executors.newSingleThreadScheduledExecutor());
 	}
 
 	@Provides
 	@Singleton
 	private VertexStore getVertexStore(
 		Universe universe,
 		RadixEngine radixEngine,
-		@Named("self") EUID self
+		@Named("self") RadixAddress self
```
talekhinezh: this can be removed

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -57,70 +64,100 @@ public EventCoordinator(
...
 			return;
 		}
 
-		this.networkSender.sendNewRound(new NewRound(round.next()));
+		try {
+			// TODO make signing more robust by including author in signed hash
+			ECDSASignature signature = this.selfKey.sign(Hash.hash256(Longs.toByteArray(view.next().number())));
+			this.networkSender.sendNewView(new NewView(selfKey.getPublicKey(), view.next(), signature));
+		} catch (CryptoException e) {
+			log.error("Failed to sign new view at " + view, e);
```
talekhinezh: This should probably be an IllegalStateException. If a node fails to sign then we don't want to continue as there are bigger problems.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -135,21 +172,13 @@ public void processProposal(Vertex proposedVertex) {
...
-					log.info("Committed vertex " + vertexId);
-
-					final Vertex vertex = vertexStore.commitVertex(vertexId);
-					final Atom committedAtom = vertex.getAtom();
-					mempool.removeCommittedAtom(committedAtom.getAID());
-				});
 		} catch (SafetyViolationException e) {
 			log.error("Rejected " + proposedVertex, e);
+		} catch (CryptoException e) {
+			log.error("Failed to vote for " + proposedVertex, e);
```
talekhinezh: Same comment as above

radixdlt/src/main/java/com/radixdlt/consensus/Vote.java
```diff
@@ -18,56 +18,77 @@
...
-import com.radixdlt.serialization.DsonOutput.Output;
 
 import java.util.Objects;
+import java.util.Optional;
 
 /**
  * Represents a vote on a vertex
  */
 @SerializerId2("consensus.vote")
+@Immutable // author cannot be but is effectively final because of serializer
```
talekhinezh, line 38: Yuck, gotta fix our serializer...

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -69,30 +113,30 @@ public void process(QuorumCertificate qc) {
 	 * @return A vote result containing the vote and any committed vertices
 	 * @throws SafetyViolationException In case the vertex would violate a safety invariant
 	 */
-	public VoteResult voteFor(Vertex proposedVertex) throws SafetyViolationException {
-		// ensure vertex does not violate earlier rounds
-		if (proposedVertex.getRound().compareTo(this.state.lastVotedRound) <= 0) {
+	public Vote voteFor(Vertex proposedVertex) throws SafetyViolationException, CryptoException {
```
talekhinezh: CryptoException here doesn't seem right.

talekhinezh (changes requested):

radixdlt/src/main/java/com/radixdlt/consensus/Vote.java
flotothemoon: Indeed..

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
flotothemoon: Yeah, that's a good idea

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
flotothemoon: Also a good idea

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
flotothemoon: yup

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
flotothemoon: Hmm, debated back and forth on this one, but happy to change it back.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
flotothemoon: Implemented as discussed

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -57,70 +64,100 @@ public EventCoordinator(
 		SafetyRules safetyRules,
 		Pacemaker pacemaker,
 		VertexStore vertexStore,
+		PendingVotes pendingVotes,
 		ProposerElection proposerElection,
-		@Named("self") EUID self
+		QuorumRequirements quorumRequirements,
```
clstrfsck: This doesn't seem to be used anywhere.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -20,47 +20,93 @@
...
 
-	@VisibleForTesting
-	Hash getCommittedVertexId(Vertex vertex) {
-		if (vertex.getRound().equals(vertex.getQC().getRound().next())
-			&& vertex.getQC().getRound().equals(vertex.getQC().getParentRound().next())) {
-			return vertex.getQC().getVertexMetadata().getParentId();
-		}
-		return null;
+	/**
+	 * Process a vertex
```
clstrfsck, line 56: First sentence of a Javadoc comment must end in a '.'

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -20,47 +20,93 @@
...
-		if (qc.getParentRound().compareTo(this.state.lockedRound) > 0) {
-			this.state = this.state.withLockedRound(qc.getParentRound());
+	public Optional<Hash> process(QuorumCertificate qc) {
+		// pre-commit phase on vertex's parent if there is a newer consecutive 1-chain
+		// keep highest 1-chain as the current "generic" QC
+		if (qc.getView().compareTo(this.state.getGenericView().orElse(View.of(0L))) > 0) {
+			this.state = this.state.withGenericQC(qc);
+		}
+
+		// commit phase on vertex's grandparent if there is a newer consecutive 2-chain
```
clstrfsck, line 76: Seems to me that this check should happen before adjusting any internal state as above.

clstrfsck: Similarly for grandparent check below.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyState.java
```diff
@@ -18,41 +18,55 @@
...
-	final Round lockedRound; // the highest 2-chain head
+	private final View lastVotedView; // the last view this node voted on (and is thus safe)
+	private final View lockedView; // the highest 2-chain head
+	private final View committedView; // the highest 3-chain (executed) head
+	private final QuorumCertificate genericQC; // the highest 1-chain head
 
 	@Inject
 	protected SafetyState() {
-		this(Round.of(0L), Round.of(0L));
+		this(View.of(0L), View.of(0L), View.of(0L), null);
```
clstrfsck: I think these should generally be `View.genesis()` instead of `View.of(0L)`

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyState.java
```diff
@@ -64,17 +78,39 @@ public boolean equals(Object o) {
 			return false;
 		}
 		SafetyState that = (SafetyState) o;
-		return Objects.equals(lastVotedRound, that.lastVotedRound)
-			&& Objects.equals(lockedRound, that.lockedRound);
+		return Objects.equals(lastVotedView, that.lastVotedView)
```
clstrfsck, line 81: Don't see any EqualsVerifier test for equals(...) and hashCode().

radixdlt/src/main/java/com/radixdlt/consensus/safety/WhitelistQuorum.java
```diff
@@ -0,0 +1,55 @@
+/*
```
clstrfsck, line 1: This seems to be a test class.  Should it move into tests?

clstrfsck (commented):
Some suggestions, but a bit concerned about the state mutation before exceptional checks in SafetyState.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyState.java
flotothemoon: Yeah, good idea

radixdlt/src/main/java/com/radixdlt/consensus/safety/WhitelistQuorum.java
flotothemoon: It's not a test class, it's a proper basic implementation of QuorumRequirements

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
flotothemoon: Indeed, originally it was used for QC aggregation but now that's handled differently. Good catch.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyState.java
flotothemoon: It's in SafetyStateTest

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
flotothemoon: Why? A one-chain can happen with the subsequent check for a two-chain failing. Same for the grandparent three-chain check below - that's the point of having these cascading checks :)

talekhinezh (approved):

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyState.java
clstrfsck: Arrgh.  Missed it 🤦‍♂

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
flotothemoon: As discussed, this is only potentially problematic under Byzantine conditions, which this PR isn't meant for anyway.

PR #40, opened 2020-03-24
Feature/validator set

clstrfsck:


radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidationResult.java
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	/**
+	 * Returns a passed result, with the specified validators signing.
+	 *
+	 * @param signed The validators who have signed
+	 * @return The validation result
+	 */
+	public static ValidationResult passed(ImmutableList<Validator> signed) {
+		return new ValidationResult(signed);
```
talekhinezh, line 61: should check that list is not empty

talekhinezh (approved):

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidationResult.java
clstrfsck: OK, fixed.

PR #43, opened 2020-03-25
Feature/multi node

talekhinezh:


radixdlt/src/main/java/com/radixdlt/consensus/EmptyEventCoordinator.java
```diff
@@ -1,38 +1,35 @@
...
+ * language governing permissions and limitations under the License.
+ */
+
+package com.radixdlt.consensus;
+
+/**
+ * An empty BFT event processor
+ */
+public class EmptyEventCoordinator implements EventCoordinator {
+	public void processVote(Vote vote) {
```
clstrfsck, line 25: All of these should have an @ Override annotation, and probably should have a comment indicating that doing nothing is on purpose, rather than an omission

radixdlt/src/main/java/com/radixdlt/consensus/BasicEpochRx.java
```diff
@@ -0,0 +1,77 @@
...
+	public BasicEpochRx(ECPublicKey selfKey, AddressBook addressBook, int fixedQuorumSize) {
+		if (fixedQuorumSize <= 0) {
+			throw new IllegalArgumentException("Quorum size must be > 0 but was " + fixedQuorumSize);
+		}
+		this.fixedQuorumSize = fixedQuorumSize;
+		this.selfKey = Objects.requireNonNull(selfKey);
+		this.addressBook = Objects.requireNonNull(addressBook);
+	}
+
+	public Observable<ValidatorSet> epochs() {
```
clstrfsck, line 54: Missing @ Override

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
```diff
@@ -0,0 +1,230 @@
...
+				final Atom committedAtom = vertex.getAtom();
+				mempool.removeCommittedAtom(committedAtom.getAID());
+			});
+
+		// proceed to next view if pacemaker feels like it
+		this.pacemaker.processQC(qc.getView())
+			.ifPresent(this::proceedToView);
+	}
+
+	public void processVote(Vote vote) {
```
clstrfsck, line 143: Bunch of methods here missing @ Override

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
```diff
@@ -32,45 +34,74 @@
...
 		bind(SafetyRules.class).in(Scopes.SINGLETON);
 	}
 
+	@Provides
+	@Singleton
+	private EpochRx epochRx(
+		@Named("self") ECKeyPair selfKey,
+		AddressBook addressBook
+	) {
+		final int fixedQuorumSize = Integer.parseInt(runtimeProperties.get("consensus.fixed_quorum_size", "1"));
```
clstrfsck, line 77: This would be better as:
```
final int fixedQuorumSize = runtimeProperties.get("consensus.fixed_quorum_size", 1);
```

clstrfsck: ~~Probably should check sensible bounds too, ie >0~~  I see this is done in the constructor.

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
```diff
@@ -32,45 +34,74 @@
...
+	@Singleton
+	private ValidatorSet validatorSet(
+		@Named("self") ECKeyPair selfKey
+	) {
+		return ValidatorSet.from(Collections.singleton(Validator.from(selfKey.getPublicKey())));
+	}
+
 	@Provides
 	@Singleton
 	private ProposerElection proposerElection(
```
clstrfsck: Should this method still be here?

radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
```diff
@@ -0,0 +1,95 @@
...
+import com.radixdlt.crypto.ECKeyPair;
+import com.radixdlt.crypto.ECPublicKey;
+import com.radixdlt.mempool.Mempool;
+import java.util.Comparator;
+import java.util.Objects;
+
+/**
+ * Manages creation of EventCoordinators given a ValidatorSet
+ */
+public class EpochManager {
```
clstrfsck, line 39: Noice.  Could probably be final.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -1,222 +1,28 @@
...
-
-	public void processLocalTimeout(View view) {
-		log.info(this.getShortName() + ": Processing LOCAL_TIMEOUT: " + view);
-
-		// proceed to next view if pacemaker feels like it
-		this.pacemaker.processLocalTimeout(view)
-			.ifPresent(this::proceedToView);
-	}
+public interface EventCoordinator {
+	void processVote(Vote vote);
```
clstrfsck, line 24: Really should have comments for these methods.  Added bonus is that they will be inherited in implementing classes.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/PacemakerImpl.java
```diff
@@ -80,28 +83,21 @@ public View getCurrentView() {
 
 	@Override
 	public Optional<View> processNewView(NewView newView) {
+		Hash newViewId = new Hash(Hash.hash256(Longs.toByteArray(newView.getView().number())));
 		ECDSASignature signature = newView.getSignature().orElseThrow(() -> new IllegalArgumentException("new-view is missing signature"));
```
clstrfsck, line 87: Would be better, IMO, to use the interfaces here, and avoid casting.  Make `pendingNewViews`:
```
private final Map<View, Signatures> pendingNewViews = new HashMap<>();
```

...and then this code could be:
```
Hash newViewId = new Hash(Hash.hash256(Longs.toByteArray(newView.getView().number())));
Signature signature = newView.getSignature().orElseThrow(() -> new IllegalArgumentException("new-view is missing signature"));
Signatures signatures = this.pendingNewViews.getOrDefault(newView.getView(), DefaultSignatures.emptySignatures());
signatures = signatures.concatenate(newView.getAuthor(), signature);
```

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
```diff
@@ -32,45 +34,74 @@
...
 		@Named("self") ECKeyPair selfKey
 	) {
-		return new Dictatorship(selfKey.getUID());
+		return new Dictatorship(selfKey.getPublicKey());
```
clstrfsck: I think Dictatorship implementation can disappear altogether now.

radixdlt/src/integration/java/com/radixdlt/consensus/MultiNodeConsensusTest.java
```diff
@@ -85,31 +83,37 @@ private ChainedBFT createBFTInstance(
...
-		PendingVotes pendingVotes = new PendingVotes(quorumRequirements);
-		EventCoordinator eventCoordinator = new EventCoordinator(
+		PacemakerImpl pacemaker = new PacemakerImpl(validatorSet, Executors.newSingleThreadScheduledExecutor());
+		PendingVotes pendingVotes = new PendingVotes();
+
+		EpochRx epochRx = () -> Observable.just(validatorSet).concatWith(Observable.never());
+		List<ECPublicKey> proposers = validatorSet.getValidators().stream()
+			.map(Validator::nodeKey)
+			.collect(Collectors.toList());
+		proposers.sort(Comparator.comparing(ECPublicKey::getUID));
```
clstrfsck, line 93: Could sort directly in the stream, but on closer inspection, it doesn't look like `proposers` is used at all, so could be removed.

radixdlt/src/main/java/com/radixdlt/consensus/EpochRx.java
```diff
@@ -1,43 +1,28 @@
...
+package com.radixdlt.consensus;
+
+import com.radixdlt.consensus.validators.ValidatorSet;
+import io.reactivex.rxjava3.core.Observable;
+
+/**
+ * Validator Set change events
+ */
+public interface EpochRx {
+	Observable<ValidatorSet> epochs();
```
clstrfsck, line 33: Documentation.

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -31,48 +31,35 @@
  * Manages pending votes for various vertices
  */
 public final class PendingVotes {
-	private final QuorumRequirements quorumRequirements;
 	private final HashMap<Hash, ECDSASignatures> pendingVotes = new HashMap<>();
```
clstrfsck, line 34: Similarly to PacemakerImpl, could be using interfaces here, and reducing amount of casting in insertVote.

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -31,48 +31,35 @@
...
 	 * @param vote The vote to be inserted
 	 * @return The generated QC, if any
 	 */
-	public Optional<QuorumCertificate> insertVote(Vote vote) {
+	public Optional<QuorumCertificate> insertVote(Vote vote, ValidatorSet validatorSet) {
 		Objects.requireNonNull(vote, "vote");
 
 		Hash voteId = vote.getVertexMetadata().getId();
 		ECDSASignature signature = vote.getSignature().orElseThrow(() -> new IllegalArgumentException("vote is missing signature"));
 		ECDSASignatures signatures = pendingVotes.getOrDefault(voteId, new ECDSASignatures());
```
clstrfsck, line 48: Should be using DefaultSignatures.emptySignatures() instead of fresh empty every call here.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/Dictatorship.java
```diff
@@ -17,28 +17,23 @@
...
 /**
  * A Proposer Election where there is only a single leader used for every view
  */
 public final class Dictatorship implements ProposerElection {
 
-	private final EUID leader;
+	private final ECPublicKey leader;
 
-	public Dictatorship(EUID leader) {
+	public Dictatorship(ECPublicKey leader) {
```
clstrfsck, line 31: This class and tests can go now I think.

clstrfsck (commented):
Nice.  Some suggestions, but nothing major.

radixdlt/src/test/java/com/radixdlt/consensus/liveness/DictatorshipTest.java
```diff
@@ -3,16 +3,15 @@
...
 public class DictatorshipTest {
 	@Test
 	public void when_retrieve_dictatorship_proposer__it_should_always_be_the_same() {
-		Dictatorship dictatorship = new Dictatorship(EUID.ONE);
-		assertThat(dictatorship.getProposer(mock(View.class))).isEqualTo(EUID.ONE);
-		assertThat(dictatorship.isValidProposer(EUID.ONE, mock(View.class))).isTrue();
-		assertThat(dictatorship.isValidProposer(mock(EUID.class), mock(View.class))).isFalse();
+		ECPublicKey ecPublicKey = mock(ECPublicKey.class);
+		Dictatorship dictatorship = new Dictatorship(ecPublicKey);
+		assertThat(dictatorship.getProposer(mock(View.class))).isEqualTo(ecPublicKey);
```
flotothemoon, line 15: Shouldn't we check this twice (i.e. call getProposer twice) to make sure it's *actually* the always the same?

radixdlt/src/test/java/com/radixdlt/consensus/EpochManagerTest.java
```diff
@@ -0,0 +1,35 @@
...
+			mock(SafetyRules.class),
+			mock(Pacemaker.class),
+			mock(VertexStore.class),
+			mock(PendingVotes.class),
+			mock(ECKeyPair.class)
+		);
+
+		ECKeyPair ecKeyPair = new ECKeyPair();
+		Validator validator = Validator.from(ecKeyPair.getPublicKey());
+		EventCoordinator eventCoordinator = epochManager.nextEpoch(ValidatorSet.from(Collections.singleton(validator)));
```
flotothemoon, line 32: This feels like a bit of a pointless test, is the failure condition really just "must not be null"?

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -36,11 +36,11 @@
  * as long as all validators sign.
  */
 public final class ValidatorSet {
-	private final Map<ECPublicKey, Validator> validators;
+	private final ImmutableBiMap<ECPublicKey, Validator> validators;
```
flotothemoon, line 39: Not clear to me why this needs to be a bi-map at this point. Is this for future use?

radixdlt/src/main/java/com/radixdlt/consensus/liveness/RotatingLeaders.java
```diff
@@ -18,26 +18,21 @@
...
 
 	@Override
-	public boolean isValidProposer(EUID nid, View view) {
-		return nid.equals(getProposer(view));
-	}
-
-	@Override
-	public EUID getProposer(View view) {
+	public ECPublicKey getProposer(View view) {
 		int index = (int) (view.number() % leaders.size());
```
flotothemoon, line 36: Appreciate that this isn't part of this PR, but I just noticed that it would be good to note that `view.number()` is guaranteed to be positive here (otherwise we might run into issues using it as an index since Java uses remainder for modulus on negative numbers).

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -31,48 +31,35 @@
...
-	public PendingVotes(QuorumRequirements quorumRequirements) {
-		this.quorumRequirements = Objects.requireNonNull(quorumRequirements);
-	}
-
 	/**
 	 * Inserts a vote for a given vertex, attempting to form a quorum certificate for that vertex.
 	 *
-	 * The vote will only be inserted if permitted by the installed {@link QuorumRequirements}.
-	 * Similarly, a QC will only be formed if permitted by the installed {@link QuorumRequirements}.
+	 * The vote will only be inserted if permitted by the {@link ValidatorSet}.
```
flotothemoon: This isn't technically true, a vote will be inserted into `signatures` even if `validatorSet` doesn't like it since there is no explicit check. The only check `validatorSet` is used for is whether a QC has formed.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinator.java
```diff
@@ -1,222 +1,28 @@
 /*
- *  (C) Copyright 2020 Radix DLT Ltd
```
flotothemoon, line 2: Looks like spurious copyright header diffs

radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
```diff
@@ -0,0 +1,95 @@
...
+	public EventCoordinator start() {
+		return new EmptyEventCoordinator();
+	}
+
+	public EventCoordinator nextEpoch(ValidatorSet validatorSet) {
+		ImmutableList<ECPublicKey> proposers = validatorSet.getValidators().stream()
+			.map(Validator::nodeKey)
+			.sorted(Comparator.comparing(ECPublicKey::getUID))
+			.collect(ImmutableList.toImmutableList());
+		ProposerElection proposerElection = new RotatingLeaders(proposers);
```
flotothemoon, line 80: It feels like this is something you should be able to configure/inject somehow, but maybe not required for now.

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
flotothemoon: Might still be useful for single node setups. But I guess those can just be represented as a multi-node setup with just one validator

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -31,48 +31,35 @@
...
-
-		// try to add the signature to form a QC if permitted by the requirements
-		if (quorumRequirements.accepts(vote.getAuthor().getUID())) {
-			// FIXME ugly cast to ECDSASignatures because we need a specific type
-			signatures = (ECDSASignatures) signatures.concatenate(vote.getAuthor(), signature);
-		} else {
-			// there is no meaningful inaction here, so better let the caller know
-			throw new IllegalArgumentException("vote " + vote + " was not accepted into QC");
-		}
+		signatures = (ECDSASignatures) signatures.concatenate(vote.getAuthor(), signature);
```
flotothemoon, line 49: Continuing from my comment above, I wonder if this blind concatenation is a potential risk. Any valid (but possibly unwanted) signature would be concatenated here without care for whether it's from a node we actually care about, potentially bottlenecking the node. Depending on which signature aggregation method we eventually used, it might also pose an actual security risk. 

The easy solution to this would be to just check if the author is included in the validator set.

flotothemoon (approved):
Generally good, some issues I would like to see addressed but nothing breaking.

radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
talekhinezh: yeah, had it as a Function parameter earlier but got a bit too messy

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
talekhinezh: yes, it's used to get set of validators and ensuring no duplicates.

radixdlt/src/test/java/com/radixdlt/consensus/EpochManagerTest.java
talekhinezh: yes, mostly to check no exceptions are thrown

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
talekhinezh: Not sure if I follow, signatures should check internally whether this signature is valid or not

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
clstrfsck: But the way this map is constructed means that it is not possible for there to be any duplicates in any case?

radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
clstrfsck: Should probably be a factory at some point.

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
clstrfsck: I agree with @flotothemoon here. If we check that `Vote.getAuthor()` is in `ValidatorsSet` before concatenating, we avoid a potential DoS / excess CPU consumption case where many invalid signatures, or valid but not part of the validator set signatures have been concatenated.

PR #49, opened 2020-03-27
Integration/responsive pacemaker

talekhinezh:


radixdlt/build.gradle
```diff
@@ -85,7 +85,7 @@ jacocoTestReport {
 dependencies {
     compile 'io.reactivex.rxjava3:rxjava:3.0.0'
     compile 'com.sleepycat:je:18.3.12'
-    compile 'com.radixdlt:radix-engine-library:1.0-beta.7'
+    compile 'com.github.radixdlt:radix-engine-library:rc~1.0-beta.8-SNAPSHOT'
```
clstrfsck: Either com.radixdlt or com.github.radixdlt will work, as we have the right TXT records in place.
https://jitpack.io/docs/#custom-domain-name

radixdlt/src/main/java/com/radixdlt/consensus/Counters.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package com.radixdlt.consensus;
+
+import java.util.EnumMap;
+import java.util.Map;
+
+/**
+ * Consensus event counting utility class.
+ * Class is NOT thread-safe so must be run in the same thread or with correct locking.
```
clstrfsck: It would be really super simple to make this thread-safe by just replacing the EnumMap with a ConcurrentMap of some kind.  I think this is really worth doing here.

radixdlt/src/integration/java/com/radixdlt/consensus/BFTTestNetwork.java
```diff
@@ -0,0 +1,138 @@
...
+import com.radixdlt.network.TestEventCoordinatorNetwork;
+import io.reactivex.rxjava3.core.Observable;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Collectors;
+
+public class BFTTestNetwork {
```
clstrfsck, line 53: I think it would help significantly if we could have a comment describing at least the rationale for these tests.

radixdlt/src/main/java/com/radixdlt/consensus/Counters.java
```diff
@@ -0,0 +1,41 @@
...
+		TIMEOUT
+	}
+
+	private Map<CounterType, Long> counters = new EnumMap<>(CounterType.class);
+
+	public void increment(CounterType counterType) {
+		counters.merge(counterType, 1L, Long::sum);
+	}
+
+	public Long getCount(CounterType counterType) {
```
clstrfsck: Probably should return primitive long.

clstrfsck (commented):
Looks OK, but would like to see Counters be made thread safe, given that this is trivial.

PR #50, opened 2020-03-27
Improvement/rpn1 61

Sajjon:


radixdlt/src/integration/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
Sajjon, line 1: This is copied over from https://github.com/radixdlt/radixdlt-core/blob/release/1.0-beta.7/radixdlt/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker

Fixes failing integrationTest due to Mockito not being able to mock final class (`AID.java`)

Sajjon:
While reviewing this you probably want to click the "gear" button in Github and enable "Hide whitespace changes", this is a one off, relating to line breaks, @clstrfsck knows more :)

Sajjon:
All `integrationTest`s, `test`s and `jmh` are passing

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
```diff
@@ -106,16 +106,13 @@ private void startQuorumNewView(View view) {
...
-			ECDSASignature signature = this.selfKey.sign(Hash.hash256(Longs.toByteArray(nextView.number())));
-			NewView newView = new NewView(selfKey.getPublicKey(), nextView, this.vertexStore.getHighestQC(), signature);
-			ECPublicKey nextLeader = this.proposerElection.getProposer(nextView);
-			log.info(String.format("%s: Sending NEW_VIEW to %s: %s", this.getShortName(), this.getShortName(nextLeader.getUID()), newView));
-			this.networkSender.sendNewView(newView, nextLeader.getUID());
-		} catch (CryptoException e) {
-			throw new IllegalStateException("Failed to sign new view", e);
-		}
+        // TODO make signing more robust by including author in signed hash
+		ECDSASignature signature = this.selfKey.sign(Hash.hash256(Longs.toByteArray(nextView.number())));
```
flotothemoon, line 109: Weird formatting

radixdlt/src/main/java/org/radix/api/services/InternalService.java
```diff
@@ -72,7 +72,7 @@ public Spammer(ECKeyPair owner, int iterations, int batching, int rate, int nonc
 			this.iterations = iterations;
 			this.rate = rate;
 			this.batching = Math.max(1, batching);
-			this.account = RadixAddress.from(universe, owner.getPublicKey());
+			this.account = new RadixAddress((byte) universe.getMagic(), owner.getPublicKey());
```
flotothemoon, line 75: Why was this changed to a constructor?

Sajjon: @flotothemoon Java Common has no knowledge of a `Universe`, hence it cannot contain that constructor.

In `Swift` we could have written that constructor inside this repo (Core), by `extension`, don't know if that is possible in Java?

flotothemoon (approved):
Not much to object to here content-wise, though there are some non-essential formatting quirks that should be addressed.

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
Sajjon: @flotothemoon yeah strange, I enabled IntelliJ to render leading whitespaces and saw the discrepancy, I believe it is fixed now.

radixdlt/src/main/java/org/radix/api/services/InternalService.java
flotothemoon: I just meant an equivalent static constructor (that doesn't need to know about universe either), that is, something like `RadixAddress.from((byte) universe.getMagic(), owner.getPublicKey())`. Up to you.

clstrfsck (approved):

PR #51, opened 2020-03-27
Integration/responsive pacemaker

talekhinezh:


radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -86,7 +86,9 @@ public void insertVertex(Vertex vertex) throws VertexInsertionException {
 			try {
 				this.engine.store(vertex.getAtom());
 			} catch (RadixEngineException e) {
-				throw new VertexInsertionException("Failed to execute", e);
+				// TODO: Don't check for state computer errors for now
```
clstrfsck: What's the though process behind this?

It seems pretty error prone to me, as it looks like ValidatingEventCoordinator will no longer remove atoms that fail to store from the memory pool.

clstrfsck (commented):
Really not sure about suppressing engine exceptions in insertVertex. Can we discuss?

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: they will get removed once the vertex is committed

radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
```diff
@@ -65,11 +69,22 @@ public void given_3_correct_bfts__then_all_should_get_same_commits_and_no_timeou
...
-			.doOnSubscribe(d -> bftNetwork.start())
+		List<Observable<Vertex>> proposals = nodes.stream()
+			.map(ECKeyPair::getUID)
+			.map(bftNetwork.getTestEventCoordinatorNetwork()::getNetworkRx)
+			.map(EventCoordinatorNetworkRx::proposalMessages)
+			.collect(Collectors.toList());
+
+		Observable<Object> proposalsCheck = Observable.merge(proposals)
+			.doOnNext(v -> assertThat(v)
+				.satisfies(new Condition<>(vtx -> vtx.getView().equals(vtx.getParentView().next()), "Vertex has direct parent")))
```
flotothemoon: We use this so often, a helper method `hasDirectParent` in `Vertex` that checks `this.getView().equals(this.getParentView().next())` might make things a little easier to read.

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
```diff
@@ -235,12 +235,11 @@ public void processProposal(Vertex proposedVertex) {
 			log.error(this.getShortName() + ": PROPOSAL: Rejected " + proposedVertex, e);
 		}
 
-		// If currently node leader, Proceed to next view
-		if (!Objects.equals(proposerElection.getProposer(updatedView), selfKey.getPublicKey())) {
+		// If not currently leader or next leader, Proceed to next view
+		if (!Objects.equals(proposerElection.getProposer(updatedView), selfKey.getPublicKey())
+			&& !Objects.equals(proposerElection.getProposer(updatedView.next()), selfKey.getPublicKey())) {
```
flotothemoon, line 235: Why does the next leader not advance to the next view here?

radixdlt/src/main/java/com/radixdlt/consensus/liveness/PacemakerImpl.java
```diff
@@ -98,14 +98,15 @@ public View getCurrentView() {
...
 
-			if (newView.getView().compareTo(this.currentView) > 0) {
+			if (newView.getView().compareTo(this.currentView) > 0
+				&& newView.getQC().getView().equals(this.currentView)) {
 				this.currentView = newView.getView();
 				scheduleTimeout(this.currentView);
 			}
 
-			if (newView.getView().compareTo(this.currentView) >= 0) {
+			if (newView.getView().compareTo(this.currentView) == 0) {
```
flotothemoon: Could just use `newView.getView().equals(this.currentView)` for clarity here

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
flotothemoon: Would be good to elaborate your reasoning in this comment here as it looks broken unless you know that

flotothemoon (approved):
Good, just some suggestions.

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
talekhinezh: since we're following chained hot stuff at the moment, we need to wait for the previous leader's QC to guarantee we have the latest QC. See the Chained HotStuff pseudocode in the paper.

flotothemoon: Ah right, since we changed back from event-driven

radixdlt/src/main/java/com/radixdlt/consensus/liveness/PacemakerImpl.java
```diff
@@ -98,14 +98,15 @@ public View getCurrentView() {
...
-			if (newView.getView().compareTo(this.currentView) > 0) {
+			if (newView.getView().compareTo(this.currentView) > 0
+				&& newView.getQC().getView().equals(this.currentView)) {
 				this.currentView = newView.getView();
 				scheduleTimeout(this.currentView);
 			}
 
-			if (newView.getView().compareTo(this.currentView) >= 0) {
+			if (newView.getView().equals(this.currentView)) {
+				pendingNewViews.remove(newView.getView());
```
flotothemoon, line 108: Shouldn't the entry in `pendingNewViews` be removed either way when a quorum is formed, regardless of whether it represents a new view?

talekhinezh: no, should still wait for previous leader's new-view. If we removed, then we would need to wait for the next timeout which we don't want.

flotothemoon: In that case comment at line 100 should be updated

PR #53, opened 2020-03-31
Bugfix/serialization

Sajjon:


radixdlt/src/main/java/org/radix/network/messaging/SignedMessage.java
```diff
@@ -20,45 +20,42 @@
...
 	// SIGNABLE //
-	public final ECDSASignature getSignature()
-	{
+	public final ECDSASignature getSignature() {
 		return this.signature;
 	}
 
-	public final void setSignature(ECDSASignature signature)
-	{
+	public final void setSignature(ECDSASignature signature) {
```
clstrfsck, line 48: I think this should be removed if you really feel like you want to do that.

flotothemoon (approved):

clstrfsck (approved):
Changes look good to me.

A minor comment about removing some dead code if you feel the need.

PR #54, opened 2020-03-31
Feature/prune vertex store

talekhinezh:


radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
```diff
@@ -83,4 +81,37 @@ public void given_4_correct_bfts__then_all_should_get_same_commits_consecutive_v
 			.take(time, timeUnit)
 			.blockingSubscribe();
 	}
+
+	@Test
+	public void given_a_correct_bft__then_new_memory_used_should_be_less_than_10MB_over_2_minutes() {
```
flotothemoon: Where does this 10MB figure come from?

radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
```diff
@@ -35,9 +35,7 @@
  */
 public class PerfectNetworkTest {
 	private static List<ECKeyPair> createNodes(int numNodes) {
```
Sajjon, line 37: Alex being annoying: but is really `numNodes` a better variable name than `nodeCount`? :D

radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
```diff
@@ -83,4 +81,37 @@ public void given_4_correct_bfts__then_all_should_get_same_commits_consecutive_v
...
+	@Test
+	public void given_a_correct_bft__then_new_memory_used_should_be_less_than_10MB_over_2_minutes() {
+		final int numNodes = 1;
+		final long time = 2;
+		final TimeUnit timeUnit = TimeUnit.MINUTES;
+
+		final List<ECKeyPair> nodes = createNodes(numNodes);
+		final BFTTestNetwork bftNetwork = new BFTTestNetwork(nodes);
+
+		Observable<Long> memory = Observable.interval(500, TimeUnit.MILLISECONDS)
```
Sajjon: `memory` -> `memoryUsedByNode`?

Sajjon (approved):
skipped through it, looks good, wrote to annoying comments about variable names, feel free to ignore :D

radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
```diff
@@ -83,4 +81,37 @@ public void given_4_correct_bfts__then_all_should_get_same_commits_consecutive_v
...
+			.map(l -> {
+				Runtime.getRuntime().gc();
```
flotothemoon: From what I remember, `Runtime.gc()` is not guaranteed to do anything before it returns (or anything at all, for that matter, since it's just a hint to the GC). So this may not work the way it was intended to.

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -43,21 +46,27 @@
 	public Optional<QuorumCertificate> insertVote(Vote vote, ValidatorSet validatorSet) {
 		Objects.requireNonNull(vote, "vote");
 
-		Hash voteId = vote.getVertexMetadata().getId();
+		Hash voteHash;
+		try {
+			voteHash = Hash.of(DefaultSerialization.getInstance().toDson(vote.getVoteData(), Output.HASH));
```
flotothemoon: I would rather inject a hasher here than have a static dependency on DefaultSerialization (which is hard to test)

radixdlt/src/main/java/com/radixdlt/consensus/VoteData.java
```diff
@@ -0,0 +1,78 @@
...
+import com.radixdlt.serialization.SerializerConstants;
+import com.radixdlt.serialization.SerializerDummy;
+import com.radixdlt.serialization.SerializerId2;
+import java.util.Objects;
+import java.util.Optional;
+import javax.annotation.concurrent.Immutable;
+
+@Immutable
+@SerializerId2("consensus.vote_data")
+public final class VoteData {
```
flotothemoon, line 38: Document me, please. Would also be useful to have an overview of the relationship between VoteData, VertexMetadata, Vertex, etc. for outsiders and future maintainers.

radixdlt/src/main/java/com/radixdlt/consensus/VoteData.java
```diff
@@ -0,0 +1,78 @@
...
+	@Override
+	public boolean equals(Object o) {
+		if (this == o) {
+			return true;
+		}
+		if (o == null || getClass() != o.getClass()) {
+			return false;
+		}
+		VoteData that = (VoteData) o;
+		return Objects.equals(proposed, that.proposed)
```
flotothemoon, line 101: Doesn't committed need to be included in equals/hashCode as well?

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -66,46 +56,38 @@ public SafetyRules(@Named("self") ECKeyPair selfKey,
...
-			// do not go beyond genesis
-			if (!grandparent.isGenesis() && twoChain) {
-				boolean threeChain = qc.getVertexMetadata().getId().equals(parent.getId())
-					&& parent.getQC().getVertexMetadata().getId().equals(grandparent.getId())
-					&& grandparent.getQC().getView().next().equals(grandparent.getView());
-				if (threeChain && grandparent.getQC().getView().compareTo(this.state.getCommittedView()) > 0) {
-					this.state = this.state.withCommittedView(grandparent.getQC().getView());
-					return Optional.of(grandparent.getQC().getVertexMetadata().getId());
-				}
+		final Optional<Hash> commitHash;
```
flotothemoon: Please document the inner workings of this, and maybe even add a note where this is from for future reference.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -127,20 +109,35 @@ public Vote voteFor(Vertex proposedVertex) throws SafetyViolationException {
...
+			committed = null;
+		}
+
+		final VoteData voteData = new VoteData(proposed, parent, committed);
 
-	@VisibleForTesting SafetyState getState() {
-		return state;
+		Hash voteHash;
+		try {
+			voteHash = Hash.of(DefaultSerialization.getInstance().toDson(voteData, Output.HASH));
```
flotothemoon: Same comment about static dependency as above

flotothemoon (changes requested):
Good changes, some concerns regarding dependencies and missing documentation.

radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
clstrfsck: I don't know what this is really testing for?  It seems like we have guessed at some values that we think are reasonable, but I don't really see why these numbers make any sense.  This is likely to make this test flakey in the long run, and the lack of explanation here for what we are trying to achieve only exacerbates that problem.

Essentially, this test is saying that 5mb/min memory consumption is OK, but clearly this is not actually OK as a steady state, as this would result in approximately 7Gb/day memory consumption.

Better would be a test that said (something like):
1. There should be a startup period, during which time it is OK for memory usage to grow.
2. After this startup period, memory growth should be, on average, zero.

radixdlt/src/main/java/com/radixdlt/consensus/VoteData.java
```diff
@@ -0,0 +1,78 @@
+package com.radixdlt.consensus;
```
clstrfsck, line 18: Missing header comment.  *Please* configure your IDE to put these in automatically for new files.  @flotothemoon or @Sajjon can help you with IntelliJ config if you need.

radixdlt/src/main/java/com/radixdlt/consensus/VoteData.java
```diff
@@ -0,0 +1,78 @@
...
+	public VertexMetadata getParent() {
+		return parent;
+	}
+
+	public Optional<VertexMetadata> getCommitted() {
+		return Optional.ofNullable(committed);
+	}
+
+	@Override
+	public boolean equals(Object o) {
```
clstrfsck, line 93: Needs an EqualsVerifier test in unit tests, given that equals is implemented here.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
clstrfsck: It really wouldn't take any significant effort to pass Serialization in as a dependency.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
clstrfsck: Might be a bit more elegant and brief as:
```
final Optional<Hash> commitHash = qc.getCommitted().flatMap(vmd -> {
	if (vmd.getView().compareTo(this.state.getCommittedView()) > 0) {
		safetyStateBuilder.committedView(vmd.getView());
		return Optional.of(vmd.getId());
	}
	return Optional.empty();
});
```

clstrfsck (changes requested):
Would really like to see EqualsVerifier tests for new classes with equals methods.

radixdlt/src/main/java/com/radixdlt/consensus/VoteData.java
talekhinezh: yup, sorry thought i had it set up but apparently not

radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
talekhinezh: All true, this was more of sanity checking where this test used to fail before this PR and now is not. Will implement the recommendation above.

clstrfsck (approved):
Looks good to me.

radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
```diff
@@ -61,26 +66,43 @@ public void given_4_correct_bfts__then_all_should_get_same_commits_consecutive_v
...
+		Observable<Object> vertexStoreCheck = Observable.timer(2, TimeUnit.SECONDS)
+			.map(l -> nodes.stream().map(bftNetwork::getVertexStore).collect(Collectors.toList()))
+			.doOnNext(s -> {
+				for (VertexStore vertexStore : s) {
+					assertThat(vertexStore.getSize()).isLessThanOrEqualTo(4);
+				}
+			})
+			.map(o -> o);
+
+		// Check that no node ever times out
```
flotothemoon, line 80: Similar to my tests over on the crash tests branch, this test cannot distinguish a dead network from a perfectly functioning one (since both will have zero timeouts). This would also benefit from a continued progress check like the one we're adding to the crash fault tests.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -101,13 +104,19 @@ public Vertex commitVertex(Hash vertexId) {
...
-		while (vertex != null && !committedVertices.containsKey(vertex.getId())) {
-			committedVertices.put(vertexId, vertex);
-			vertex = vertices.get(vertex.getParentId());
+		while (vertex != null && !root.equals(vertex)) {
+			path.addFirst(vertex);
+			vertex = vertices.remove(vertex.getParentId());
+		}
+
+		for (Vertex committed : path) {
+			lastCommittedVertex.onNext(committed);
```
flotothemoon, line 115: Since there are no checks whether `committed` was already committed by an earlier call to `commitVertex` it's likely that there will be many duplicates in lastCommittedVertex. I think that's odd, but if it's intended behaviour, it should be documented in the lastCommittedVertex method.

radixdlt/src/main/java/com/radixdlt/consensus/VoteData.java
```diff
@@ -0,0 +1,102 @@
...
+		this.proposed = Objects.requireNonNull(proposed);
+		this.parent = parent;
+		this.committed = committed;
+	}
+
+	public VertexMetadata getProposed() {
+		return proposed;
+	}
+
+	public VertexMetadata getParent() {
```
flotothemoon, line 84: It seems inconsistent to have getParent() be a nullable getter and getCommitted() be an Optional.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -66,46 +60,32 @@ public SafetyRules(@Named("self") ECKeyPair selfKey,
 	 * @return the just-committed vertex id, if any
 	 */
 	public Optional<Hash> process(QuorumCertificate qc) {
+		final Builder safetyStateBuilder = this.state.toBuilder();
```
flotothemoon, line 63: Nice, good idea.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -66,46 +60,32 @@ public SafetyRules(@Named("self") ECKeyPair selfKey,
...
-					return Optional.of(grandparent.getQC().getVertexMetadata().getId());
-				}
+		final Optional<Hash> commitHash = qc.getCommitted().flatMap(vmd -> {
+			if (vmd.getView().compareTo(this.state.getCommittedView()) > 0) {
+				safetyStateBuilder.committedView(vmd.getView());
+				return Optional.of(vmd.getId());
 			}
-		}
+			return Optional.empty();
+		});
```
flotothemoon: Would really like to see some documentation for these lines considering how critical they are. Even just a reference to the relevant section in the HS paper would be great.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -66,46 +60,32 @@ public SafetyRules(@Named("self") ECKeyPair selfKey,
 	 * @return the just-committed vertex id, if any
 	 */
 	public Optional<Hash> process(QuorumCertificate qc) {
+		final Builder safetyStateBuilder = this.state.toBuilder();
+
 		// pre-commit phase on vertex's parent if there is a newer consecutive 1-chain
```
flotothemoon: This doesn't actually check for consecutiveness, right?

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -127,20 +107,29 @@ public Vote voteFor(Vertex proposedVertex) throws SafetyViolationException {
...
+		final VertexMetadata parent = VertexMetadata.ofParent(proposedVertex);
+		final VertexMetadata committed;
 
-	@VisibleForTesting SafetyState getState() {
-		return state;
+		if (proposedVertex.getView().equals(proposedVertex.getParentView().next())
+			&& !proposedVertex.getParentView().isGenesis() && !proposedVertex.getGrandParentView().isGenesis()
+			&& proposedVertex.getParentView().equals(proposedVertex.getGrandParentView().next())
+		) {
+			committed = proposedVertex.getQC().getParent();
```
flotothemoon: At this point we're at round r+2 for `committed`, so it would technically be `toBeCommitted` since it's what *will* be committed when r+2 is certified in r+3

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyState.java
```diff
@@ -45,24 +45,54 @@ protected SafetyState() {
 		this.genericQC = genericQC;
 	}
 
-	SafetyState(SafetyState other) {
-		this(other.lastVotedView, other.lockedView, other.committedView, other.genericQC);
-	}
+	public static class Builder {
```
flotothemoon: Does this need to be public?

radixdlt/src/test/java/com/radixdlt/consensus/VertexStoreTest.java
```diff
@@ -114,5 +113,26 @@ public void when_insert_and_commit_vertex__then_committed_vertex_should_emit() t
 		assertThat(vertexStore.commitVertex(nextVertex.getId())).isEqualTo(nextVertex);
 		testObserver.awaitCount(2);
 		testObserver.assertValues(genesisVertex, nextVertex);
+		assertThat(vertexStore.getSize()).isEqualTo(1);
+	}
+
+	@Test
+	public void when_insert_two_and_commit_vertex__then_two_committed_vertices_should_emit_in_order_and_store_should_have_size_1() throws Exception {
```
flotothemoon, line 139: It would be good to have a test case for when `nextVertex` is inserted and committed and *then* `nextVertex2` is inserted and committed since that should be well-defined behaviour.

radixdlt/src/test/java/com/radixdlt/consensus/safety/SafetyRulesTest.java
```diff
@@ -40,171 +40,126 @@
...
-		safetyRules.process(b2);
-		assertThatThrownBy(() -> safetyRules.voteFor(b2));
+		Vote vote = safetyRules.voteFor(proposal);
+		assertThat(vote.getVoteData().getCommitted()).hasValue(VertexMetadata.ofVertex(grandParent));
+	}
 
-		assertThat(safetyRules.process(a3)).isEmpty();
-		safetyRules.voteFor(a3);
+	@Test
+	public void when_vote_on_proposal_three_after_genesis_with_skip__then_returned_vote_has_no_commit() throws SafetyViolationException {
```
flotothemoon, line 127: So for now, we're requiring consecutiveness for our chains. Would be good to document this somewhere (e.g. in SafetyRules).

flotothemoon (changes requested):
Good work, there remain only some smaller concerns that I would like to see addressed considering how critical this code is.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: There shouldn't be any duplicates given the `!root.equals(vertex)` check in the loop. This is also tested for in the unit test.

radixdlt/src/integration/java/com/radixdlt/consensus/PerfectNetworkTest.java
talekhinezh: True, but probably not related to this PR. Will add a task for adding that.

flotothemoon (approved):
Fine by me, well done.

PR #55, opened 2020-03-31
Add (more) crash fault tests

flotothemoon:
Adds two new tests to the `CrashFaultNetworkTest.java` integration test suite to exercise the following configurations (in total):
1. 1 out of 3 nodes is crashed, should not make progress because QC cannot be formed
2. 1 out of 4 nodes is crashed, should make progress whenever a correct instance is leader 
3. 7 correct nodes, randomly crashing correct nodes, should make progress until QCs are impossible

To enable these tests I also expanded the capabilities of our test suite and made minor utility modifications in core code.


radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
```diff
@@ -17,52 +17,203 @@
...
+import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
 
 /**
- * Tests with networks with crashed nodes
+ * Tests with networks with crash-stopped nodes (that do not recover).
+ * These tests comprise both static and dynamic configurations (where nodes crash before or during runtime, respectively).
  */
 public class CrashFaultNetworkTest {
-	static List<ECKeyPair> createNodes(int numNodes) {
+	static List<ECKeyPair> createNodes(int numNodes, String designation) {
```
flotothemoon: This isn't the most beautiful way to do this, but it's fast enough and very useful for debugging. However, it only works for static configurations (since a node short name does not change over time). We should consider adding the ability to set a custom name to BFT instances, if just for these kind of tests (e.g. to mark them as "faulty 1" or whatever).

flotothemoon:
There appears to be a failing PacemakerImplTest, looking into that now.

radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
```diff
@@ -17,52 +17,203 @@
...
 		return Stream.generate(() -> {
-			return ECKeyPair.generateNew();
+			// efficient? no. useful? yes.
+			ECKeyPair keyPair;
+			do {
+				keyPair = ECKeyPair.generateNew();
+			} while (!keyPair.euid().toString().startsWith(designation));
```
talekhinezh: POW?!

flotothemoon: Essentially, yes, though with a *super* low "difficulty" and for a good reason (interpretability of results and logs). I would like a more proper feature for this but we don't have one yet, so it's a placeholder.

radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
```diff
@@ -17,52 +17,203 @@
...
+			.map(committedVertices -> committedVertices.distinct().collect(Collectors.toList()))
+			.doOnNext(committedVertices -> assertThat(committedVertices).hasSize(1))
+			.map(vertices -> (Vertex) vertices.get(0))
+			.map(o -> o);
+
+		// correct nodes should only get timeouts when crashed nodes were a proposer
+		Observable<Object> correctTimeoutCheck = Observable.interval(1, TimeUnit.SECONDS)
+			.flatMapIterable(i -> correctNodes)
+			.doOnNext(cn -> assertThat(bftNetwork.getCounters(cn).getCount(Counters.CounterType.TIMEOUT))
+				.satisfies(new Condition<>(c -> c == (bftNetwork.getPacemaker(cn).getCurrentView().number() / numNodes) * numCrashed,
```
talekhinezh: Aren't there race conditions here?

flotothemoon: How so? The number of crashed nodes is static here

radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
```diff
@@ -17,52 +17,203 @@
...
+				.map(nodeAndVertex -> (Pair<ECPublicKey, Vertex>) nodeAndVertex)
+				.filter(nodeAndVertex -> !faultyNodesPubs.contains(nodeAndVertex.getFirst()))
+				.map(Pair::getSecond)
+				.distinct()
+				.collect(Collectors.toList()))
+			.doOnNext(committedVertices -> assertThat(committedVertices).hasSize(1))
+			.map(vertices -> vertices.get(0))
+			.map(o -> o);
+		//CHECKSTYLE:ON
+
```
talekhinezh, line 225: Would like a check to make sure that bft is always making progress here

flotothemoon: Good idea

radixdlt/src/main/java/com/radixdlt/consensus/ProposerElectionFactory.java
```diff
@@ -0,0 +1,27 @@
...
+
+package com.radixdlt.consensus;
+
+import com.radixdlt.consensus.liveness.ProposerElection;
+import com.radixdlt.crypto.ECPublicKey;
+
+import java.util.Collection;
+
+public interface ProposerElectionFactory {
+	ProposerElection create(Collection<ECPublicKey> proposers);
```
talekhinezh: This should be an ImmutableList since order of proposers will matter given that proposers is something serialized and sent over the network.

flotothemoon: Good point on the list, will change that. Would rather avoid bringing ImmutableList into this, can always just make an immutable copy internally (that is, I think it's fine to put this requirement into documentation)

radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
talekhinezh: if for example pacemaker timesout in between calls `getCount` and `getCurrentVew()`

flotothemoon: Ah yes, I figured the probability of that happening was super low. Do you think it matters?

talekhinezh: Don't think the probability of that is that low and false negatives here would just make things confuse people down the line. Can you think of a cleaner way of doing this check? Maybe a counter per node leader timeout instead of a global one?

radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
flotothemoon: After some more changes this isn't that useful anymore, so I've removed it for now

radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
flotothemoon: Possibly, I'll think about it.

flotothemoon: The simplest I've come up with is just checking that the number of actual timeouts is *smaller* than the maximum number of possible timeouts. In case of a race condition where the pacemaker times out between calls this still works.

radixdlt/src/main/java/com/radixdlt/consensus/ProposerElectionFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.radixdlt.consensus;
+
+import com.radixdlt.consensus.liveness.ProposerElection;
+import com.radixdlt.crypto.ECPublicKey;
+
+import java.util.List;
+
+public interface ProposerElectionFactory {
```
talekhinezh, line 28: javadoc

flotothemoon: argh! added, thanks.

PR #60, opened 2020-04-03
Feature/latent msgs

talekhinezh:
* Fix bug with latent messages (e.g. new-view from previous leader coming before everyone else)

### DIFF
--- a/radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
+++ b/radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
@@ -175,7 +175,7 @@ public final class ValidatingEventCoordinator implements EventCoordinator {
 			return;
 		}
 
-		this.pacemaker.processNewView(newView, validatorSet, proposerElection)
+		this.pacemaker.processNewView(newView, validatorSet)
 			.ifPresent(syncedView -> {
 				// Hotstuff's Event-Driven OnBeat
 				Vertex proposal = proposalGenerator.generateProposal(syncedView);
--- a/radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
+++ b/radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
@@ -94,7 +94,7 @@ public final class ValidatingEventCoordinator implements EventCoordinator {
 		return getShortName(selfKey.euid());
 	}
 
-	// OnNextSyncView
+	// Hotstuff's Event-Driven OnNextSyncView
 	private void proceedToView(View nextView) {
 		// TODO make signing more robust by including author in signed hash
 		ECDSASignature signature = this.selfKey.sign(Hash.hash256(Longs.toByteArray(nextView.number())));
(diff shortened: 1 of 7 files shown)


radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
```diff
@@ -94,17 +94,7 @@ private String getShortName() {
...
-		if (!Objects.equals(proposerElection.getProposer(view), selfKey.getPublicKey())) {
-			return;
-		}
-
-		Vertex proposal = proposalGenerator.generateProposal(view);
-		log.info("{}: Broadcasting PROPOSAL: {}", getShortName(), proposal);
-		this.networkSender.broadcastProposal(proposal);
-	}
-
+	// OnNextSyncView
```
flotothemoon: I presume this refers to the event-driven HotStuff described in the original paper. If so, maybe we can explicitly state that, cause I'm not sure I remember what this refers to a few months from now.

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
```diff
@@ -179,8 +169,13 @@ public void processNewView(NewView newView) {
 			return;
 		}
 
-		this.pacemaker.processNewView(newView, validatorSet)
-			.ifPresent(this::startQuorumNewView);
+		this.pacemaker.processNewView(newView, validatorSet, proposerElection)
+			.ifPresent(syncedView -> {
+				// OnBeat
```
flotothemoon: Same here as above (regarding HS paper reference).

radixdlt/src/main/java/com/radixdlt/consensus/liveness/PacemakerImpl.java
```diff
@@ -90,34 +90,39 @@ public View getCurrentView() {
...
-			if (newView.getView().equals(this.currentView)) {
-				pendingNewViews.remove(newView.getView());
-				return Optional.of(this.currentView);
-			} else {
-				log.info("Ignoring New View Quorum: " + newView.getView() + " Current is: " + this.currentView);
+		// TODO: Should check that previous leader is correct by checking new-view has qc for his view
+		final boolean isFromPrevLeader = Objects.equals(proposerElection.getProposer(newView.getView().previous()), newView.getAuthor())
+			&& !newView.getQC().getView().isGenesis();
+
+		if (!isFromPrevLeader) {
```
flotothemoon: It's not clear to me why the previous leader is excluded from the quorum, could you please explain?

radixdlt/src/test/java/com/radixdlt/consensus/liveness/PacemakerImplTest.java
```diff
@@ -185,41 +187,52 @@ public void when_inserting_valid_but_old_new_views__then_no_new_view_is_returned
...
 		when(validatorSet.validate(any(), any())).thenReturn(result);
 		ScheduledExecutorService executorService = getMockedExecutorService();
 		PacemakerImpl pacemaker = new PacemakerImpl(TEST_PACEMAKER_TIMEOUT, executorService);
-		assertThat(pacemaker.processNewView(newView, validatorSet)).isPresent().get().isEqualTo(View.of(0));
-		assertThat(pacemaker.getCurrentView()).isEqualTo(View.of(0));
-		verify(executorService, times(0)).schedule(any(Runnable.class), anyLong(), any());
+		pacemaker.processQC(View.of(0));
+
+		assertThat(pacemaker.processNewView(newView, validatorSet, mock(ProposerElection.class))).isPresent().get().isEqualTo(View.of(1));
+		assertThat(pacemaker.getCurrentView()).isEqualTo(View.of(1));
```
flotothemoon, line 205: The method is called `when_inserting_current_and_accepted_new_views__then_qc_is_formed_and_current_view_has_changed_and_no_new_timeout` but it doesn't actually test for new timeouts

flotothemoon (commented):
Some preliminary comments pending further clarification.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/PacemakerImpl.java
talekhinezh: if you get the previous leader's new-view with prev view's QC, it's proof that a quorum has already moved onto the next view. But I think perhaps this is only possible if currently in the previous view.

PR #62, opened 2020-04-06
Feature/rpnv1 128 integration tests on travis

clstrfsck:
Add integration test step to travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ jobs:
     - stage: sonar
       script: ./gradlew jacocoTestReport sonarqube --info
     - stage: integration
-      script: ./gradlew integrationTest --info
+      script: ./gradlew integrationTest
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,6 +20,8 @@ jobs:
       script: ./gradlew check --info
     - stage: sonar
       script: ./gradlew jacocoTestReport sonarqube --info
+    - stage: integration
+      script: ./gradlew integrationTest
```
clstrfsck, line 24: Run the integration tests last, so that sonar results come out quicker, as the integration tests take 10 mins or so to complete.

Run the integration tests without `--info`, as the logging is very verbose, and sometimes exceeds the amount of output that Travis allows, resulting in the job being terminated.

talekhinezh (approved):

PR #63, opened 2020-04-07
Add optional initial heap filling and heap dump on OOMError

clstrfsck:


radixdlt/src/main/java/org/radix/Radix.java
```diff
@@ -71,6 +77,22 @@
 	private static final Object BC_LOCK = new Object();
 	private static boolean bcInitialised;
 
+	// Be careful here.  This is a debugging method to detect overcommitted heap usage on Linux.
```
flotothemoon: It's not quite clear to me what this method is used for. Is this just for debugging?

clstrfsck: It's used to attempt to determine if the heap size is overcommitted on Linux (as per the comment).

Linux will allow processes to overcommit memory so, for example, a JVM can ask for 2Gb of heap on a Linux machine that only has, for example, 1Gb of physical memory.  This request will succeed.

Once the process "touches" the virtual memory allocated, it will be assigned to physical memory.  If this assignment fails because there is no physical memory available to assign, then the process is killed by the Linux Out Of Memory Killer (also known as the OOM Killer).

Because of concurrent GC, it's unlikely that our Java processes will reach anything like maximum heap usage under normal/quiescent loads, ie on startup.  Unfortunately it's only likely that we will reach max heap usage under extreme conditions, ie there is limited CPU to allocate to GC tasks, which usually happens much later in the processes life, and is somewhat easily confused with memory leaks.

This method is an attempt to touch at least most of the pages in the heap on startup so that any mismatch between the heap requested and the available memory on the underlying OS/container is caught early.

flotothemoon: I see. This process then presumes that the memory commitments don't change, that is, that after the time this "test" is performed, other applications don't claim the memory we immediately free again. My understanding of how Linux memory management and the JVM interact is fuzzy though, so maybe this isn't an issue.

clstrfsck: There are two layers of memory management happening here.  One at the OS level (JVM asks OS for certain amount of memory for heap etc), and one at the JVM level (application and runtime makes requests from the JVM).  It's the interaction between the two that can and has caused us issues.  However, freeing/garbage collecting at the JVM level has no effect on OS allocations.

From the Linux/OS point of view, the memory is never freed/unmapped once it is allocated/mapped to physical memory (ignoring swap which we don't use).  It can't be freed, otherwise the contents would be lost.

As it turns out, there are ways of disabling memory overcommit on Linux, and we should explore these for docker containers.

talekhinezh (approved):

PR #66, opened 2020-04-10
Bugfix/fix node count config name

talekhinezh:


radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
```diff
@@ -20,6 +20,7 @@
 import com.radixdlt.crypto.ECKeyPair;
 import com.radixdlt.crypto.ECPublicKey;
 import com.radixdlt.examples.tictactoe.Pair;
```
clstrfsck: Oh, that's not great (import from examples).  I didn't notice that before.

flotothemoon (approved):

radixdlt/src/main/java/org/radix/Radix.java
```diff
@@ -71,6 +77,22 @@
 	private static final Object BC_LOCK = new Object();
 	private static boolean bcInitialised;
 
+	// Be careful here.  This is a debugging method to detect overcommitted heap usage on Linux.
```
clstrfsck: This should not be here.  Need to rebase on current head of `feature/rpnv1-67-oom-killer-linux`

clstrfsck (changes requested):

radixdlt/src/integration/java/com/radixdlt/consensus/CrashFaultNetworkTest.java
talekhinezh: confused why this file is even showing up here

clstrfsck (approved):
Looks good when we resolve the conflict.

PR #67, opened 2020-04-10
Environment variable for pacemaker timeout

shambupujar:
Change to enable setting up pacemaker timeout through environment variable


flotothemoon (approved):

flotothemoon (approved):

docker/config/default.config.envsubst
```diff
@@ -24,6 +24,7 @@ partition.fragments=$RADIXDLT_PARTITION_FRAGMENTS
 universe.witness=$RADIXDLT_UNIVERSE_WITNESSES
 debug.profiler=$RADIXDLT_DEBUG_PROFILER
 consensus.fixed_quorum_size=$RADIXDLT_CONSENSUS_FIXED_QUORUM_SIZE
+consensus.pacemaker_timeout_millis=$RADIXDLT_PACEMAKER_TIMEOUT_MILLIS
```
clstrfsck, line 27: I think you will need to set this environment variable in (at least) Dockerfile as well, otherwise we may end up with a timeout of 0.

clstrfsck (changes requested):

docker/config/default.config.envsubst
shambupujar: Have setup default value of 5000ms in Dockerfile

talekhinezh (approved):

clstrfsck (approved):

PR #68, opened 2020-04-13
Bugfix/process msgs in order

talekhinezh:
Ensure that consensus messages are processed in the order they are received


radixdlt/src/main/java/com/radixdlt/network/ConsensusMessageDto.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * The Data Transfer Object for Consensus messages. Each type of consensus message currently needs to be
+ * a parameter in this class. Though "ugly" it is required due to:
+ * 1. the current requirement of consensus to process messages in order
+ * 2. the current interface design of MessageCentral which requires listeners to
+ * register with a single class, otherwise in-order messages isn't guaranteed
+ */
+@SerializerId2("message.consensus.msg")
+public class ConsensusMessageDto extends Message {
```
clstrfsck: This class has too many suffixes.  Can we please just call it ConsensusMessage.

radixdlt/src/main/java/com/radixdlt/network/ConsensusMessageDto.java
```diff
@@ -0,0 +1,84 @@
...
+
+	ConsensusMessageDto(int magic, Vote vote) {
+		super(magic);
+		this.newView = null;
+		this.proposal = null;
+		this.vote = vote;
+	}
+
+	public ConsensusMessage getConsensusMessage() {
+		return this.newView != null ? this.newView : this.proposal != null
```
clstrfsck: Should not be using nested ternary operators.

radixdlt/src/main/java/com/radixdlt/consensus/ChainedBFT.java
```diff
@@ -96,7 +96,7 @@ public ChainedBFT(
 			.startWithItem(epochManager.start())
 			.doOnNext(EventCoordinator::start)
 			.publish()
-			.autoConnect(4);
+			.autoConnect(2);
```
clstrfsck: Constant here seems error prone.

radixdlt/src/main/java/com/radixdlt/consensus/ChainedBFT.java
```diff
@@ -107,35 +107,28 @@ public ChainedBFT(
...
+		final Observable<Event> consensusMessages = Observable.combineLatest(
 			epochCoordinators,
-			this.network.proposalMessages(),
-			(e, proposal) -> {
-				e.processProposal(proposal);
-				return new Event(EventType.PROPOSAL_MESSAGE, proposal);
+			this.network.consensusMessages(),
+			(e, msg) -> {
+				final EventType eventType;
+				if (msg instanceof NewView) {
```
clstrfsck, line 115: Might be work considering a Map<Class, Consumer> for dispatch, instead of this chain of if/else.

radixdlt/src/main/java/com/radixdlt/consensus/ConsensusMessage.java
```diff
@@ -0,0 +1,26 @@
...
+ * language governing permissions and limitations under the License.
+ */
+
+package com.radixdlt.consensus;
+
+/**
+ * A message meant for consensus
+ * TODO: add signatures and validation here
+ */
+public interface ConsensusMessage {
```
clstrfsck: Is this intended to be a marker interface (TODO suggests not)? 

If so, do we really need this?

clstrfsck: Also would suggest using a suffix other that "Message", as "Message" suffix is currently used exclusively by things that derive from the Message class.

radixdlt/src/main/java/com/radixdlt/consensus/EventCoordinatorNetworkSender.java
```diff
@@ -17,13 +17,13 @@
...
+import com.radixdlt.crypto.ECPublicKey;
 
 /**
  * Interface for Event Coordinator to send things through a network
  */
 public interface EventCoordinatorNetworkSender {
-	void broadcastProposal(Vertex vertex);
-	void sendNewView(NewView newView, EUID newViewLeader);
-	void sendVote(Vote vote, EUID leader);
+	void broadcastProposal(Proposal proposal);
```
clstrfsck, line 31: Methods should be documented.

radixdlt/src/main/java/com/radixdlt/network/ConsensusMessageDto.java
```diff
@@ -0,0 +1,84 @@
...
+import com.radixdlt.serialization.DsonOutput;
+import com.radixdlt.serialization.DsonOutput.Output;
+import com.radixdlt.serialization.SerializerId2;
+import org.radix.network.messaging.Message;
+
+/**
+ * The Data Transfer Object for Consensus messages. Each type of consensus message currently needs to be
+ * a parameter in this class. Though "ugly" it is required due to:
+ * 1. the current requirement of consensus to process messages in order
+ * 2. the current interface design of MessageCentral which requires listeners to
```
clstrfsck: As discussed, I'll make the inbound / outbound queue processing single threaded.

radixdlt/src/main/java/com/radixdlt/consensus/ConsensusMessage.java
clstrfsck: Also, why is this not a (possibly) abstract class deriving from Message?

clstrfsck (changes requested):
Can we fix up the ConsensusMessage / ConsensusMessageDto stuff to be more consistent with existing class naming?

Otherwise some minor issues, and not too clear on the purpose of the ConsensusMessage interface.

radixdlt/src/main/java/com/radixdlt/consensus/ChainedBFT.java
talekhinezh: the constant is dependent on number of subscribers which is only known at pre-compile time, not sure if there's a clearer way of implementing this.

radixdlt/src/main/java/com/radixdlt/consensus/ConsensusMessage.java
talekhinezh: Would really like to move away from DTO level objects being utilized within functional Consensus logic.

radixdlt/src/main/java/com/radixdlt/consensus/ChainedBFT.java
talekhinezh: that sounds even more nasty as `Consumer` itself can't be typed. Given the nastiness around that, seems to me this way is currently the least of the two evils.

radixdlt/src/main/java/com/radixdlt/network/ConsensusMessageDto.java
talekhinezh: Problem is there are two levels of "messages", a network message and a consensus functional message.

clstrfsck (approved):
Looks good.

PR #71, opened 2020-04-15
Added logging and toString methods to support network tracing

clstrfsck:
Added logging to feature/tcp-by-default branch


radixdlt/src/main/java/com/radixdlt/consensus/Proposal.java
```diff
@@ -79,8 +79,11 @@ private void setSerializerAuthor(byte[] author) throws CryptoException {
 
 	@Override
 	public String toString() {
-		return String.format("%s{author=%s view=%s}", getClass().getSimpleName(),
-			author.euid().toString().substring(0, 6), vertex.getView());
+		String who = author == null ? "null" : author.euid().toString().substring(0, 6);
```
flotothemoon: Why make null a string here (but not for the other properties)?

radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
```diff
@@ -76,11 +75,11 @@
 	private final RateLimiter outboundLogRateLimiter = RateLimiter.create(1.0);
 
 	// Inbound message handling
-	private final BlockingQueue<MessageEvent> inboundQueue;
+	private final SimpleBlockingQueue<MessageEvent> inboundQueue;
 	private final SimpleThreadPool<MessageEvent> inboundThreadPool;
 
 	// Outbound message handling
-	private final BlockingQueue<MessageEvent> outboundQueue;
+	private final SimpleBlockingQueue<MessageEvent> outboundQueue;
```
flotothemoon, line 82: It's not evident to me why this was changed, could you elaborate?

radixdlt/src/test/java/com/radixdlt/network/messaging/MessageDispatcherTest.java
```diff
@@ -168,21 +162,41 @@ public void receiveExpiredMessage() {
...
     }
 
+    @Test
+    public void receiveMessageFromBannedPeer() {
+    	// Banned for a long time
+    	peer1.setBan("Test", System.currentTimeMillis() + 86_400_000L);
+    	when(this.addressBook.peer(any(EUID.class))).thenReturn(Optional.of(peer1));
+    	RadixSystem system = mock(RadixSystem.class);
+    	doReturn(EUID.TWO).when(system).getNID();
+    	doReturn(Radix.AGENT_VERSION).when(system).getAgentVersion();
```
talekhinezh: weird tabbing

talekhinezh (approved):

radixdlt/src/main/java/com/radixdlt/network/messaging/MessageDispatcher.java
```diff
@@ -213,4 +222,31 @@ private SendResult updateStatistics(SendResult result) {
...
+	 * known peer's banned time.
+	 *
+	 * @param peerNid the corresponding node ID of the peer
+	 * @param messageType the message type for logging ignored messages
+	 * @return {@code true} if the peer is currently banned, {@code false} otherwise
+	 */
+	private boolean checkPeerBanned(EUID peerNid, String messageType) {
+		return this.addressBook.peer(peerNid)
+			.filter(kp -> kp.getTimestamp(Timestamps.BANNED) > this.timeSource.currentTime())
+			.map(kp -> {
```
flotothemoon, line 246: This seems like a counter-intuitive place to do the actual logging, I would expect that to be above the part that returns in the calling method.

flotothemoon (approved):

radixdlt/src/main/java/com/radixdlt/consensus/Proposal.java
clstrfsck: The others aren't strings, so a string wouldn't work there, but an actual `null` would be fine here, so I'll change it to that.

radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
clstrfsck: Primarily because this was previously a `PriorityBlockingQueue`.  Unfortunately the Javadoc for that class states "_Operations on this class make no guarantees about the ordering of elements with equal priority_", which completely breaks the LIFO behaviour we need to guarantee consensus messages are sent and received in order.

radixdlt/src/main/java/com/radixdlt/network/messaging/MessageDispatcher.java
clstrfsck: Yeah, I know.  I wanted to include the existing peer in the log message, and returning it elsewhere to allow logging to be more consistent seemed worse than the inconsistency to me.

radixdlt/src/test/java/com/radixdlt/network/messaging/MessageDispatcherTest.java
clstrfsck: Actually, this whole file seems to use The Correct Indentation Method, rather than tabs.  Will uncorrect it.

PR #72, opened 2020-04-20
Refactor latent network test

talekhinezh:
* Increase variance of latency to catch "node-falling-behind" problem

### DIFF
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
@@ -25,7 +25,10 @@ import io.reactivex.rxjava3.core.Observable;
 public interface BFTCheck {
 
 	/**
-	 * Creates an observable which runs assertions against a bft network
+	 * Creates an observable which runs assertions against a bft network.
+	 * Assertions errors are expected to propagate down the observable.
+	 * TODO: Cleanup interface a bit
+	 *
 	 * @param network network to check
 	 * @return observable to subscribe to enable checking
 	 */
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
@@ -1,0 +1,24 @@
+/*
+ * (C) Copyright 2020 Radix DLT Ltd
+ *
+ * Radix DLT Ltd licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except in
+ * compliance with the License.  You may obtain a copy of the
+ * License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND,
+ * either express or implied.  See the License for the specific
+ * language governing permissions and limitations under the License.
+ */
+
+package com.radixdlt.consensus;
+
+import io.reactivex.rxjava3.core.Observable;
+
+public interface BFTCheck {
+	Observable<Object> check(BFTTestNetwork network);
+}
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
@@ -19,6 +19,15 @@ package com.radixdlt.consensus;
 
 import io.reactivex.rxjava3.core.Observable;
 
+/**
+ * A running BFT check given access to network
+ */
 public interface BFTCheck {
+
+	/**
+	 * Creates an observable which runs assertions against a bft network
+	 * @param network network to check
+	 * @return observable to subscribe to enable checking
+	 */
 	Observable<Object> check(BFTTestNetwork network);
 }
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
@@ -18,7 +18,7 @@
 package com.radixdlt.consensus;
 
 import com.radixdlt.consensus.checks.AllProposalsHaveDirectParentsCheck;
-import com.radixdlt.consensus.checks.LivenessCheck;
+import com.radixdlt.consensus.checks.StrongLivenessCheck;
 import com.radixdlt.consensus.checks.NoSyncExceptionCheck;
 import com.radixdlt.consensus.checks.NoTimeoutCheck;
 import com.radixdlt.consensus.checks.SafetyCheck;
@@ -85,7 +85,7 @@ public class BFTTest {
 	}
 
 	public void assertLiveness() {
-		this.checks.add(new LivenessCheck());
+		this.checks.add(new StrongLivenessCheck());
 	}
 
 	public void assertSafety() {
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
@@ -17,26 +17,23 @@
 
 package com.radixdlt.consensus;
 
-import static org.assertj.core.api.AssertionsForInterfaceTypes.assertThat;
-
-import com.radixdlt.counters.SystemCounters.CounterType;
+import com.radixdlt.consensus.checks.AllProposalsHaveDirectParentsCheck;
+import com.radixdlt.consensus.checks.LivenessCheck;
+import com.radixdlt.consensus.checks.NoSyncExceptionCheck;
+import com.radixdlt.consensus.checks.NoTimeoutCheck;
+import com.radixdlt.consensus.checks.SafetyCheck;
 import com.radixdlt.crypto.ECKeyPair;
 import com.radixdlt.middleware2.network.TestEventCoordinatorNetwork;
 import io.reactivex.rxjava3.core.Observable;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.assertj.core.api.AssertionsForClassTypes;
-import org.assertj.core.api.Condition;
 
 public class BFTTest {
 	private final long time;
 	private final TimeUnit timeUnit;
-	private final List<ECKeyPair> allNodes;
 	private final BFTTestNetwork bftNetwork;
 	private final List<Observable<Object>> assertions = new ArrayList<>();
 	private static final int MINIMUM_NETWORK_LATENCY = 10;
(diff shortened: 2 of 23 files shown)


radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
```diff
@@ -0,0 +1,113 @@
...
+import io.reactivex.rxjava3.core.Observable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class BFTTest {
+	private final int numNodes;
+	private final long time;
```
flotothemoon, line 36: It seems cleaner to me to extract the configuration properties (i.e. everything except `checks`) into some BFTTestConfiguration instance and have a builder for that, which would enable us to test the same checks against various configurations more easily. Might be premature, that's for you to judge :)

radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package com.radixdlt.consensus;
+
+import io.reactivex.rxjava3.core.Observable;
+
+/**
+ * A running BFT check given access to network
+ */
+public interface BFTCheck {
```
flotothemoon, line 25: Great idea! I'm quite fond of this architecture, makes things a lot cleaner.

radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
```diff
@@ -0,0 +1,33 @@
...
+
+	/**
+	 * Creates an observable which runs assertions against a bft network
+	 * @param network network to check
+	 * @return observable to subscribe to enable checking
+	 */
+	Observable<Object> check(BFTTestNetwork network);
```
flotothemoon, line 35: Just going by the documentation here it's not 100% clear to me what is supposed to happen if an assertion fails. Throw an assertion exception? Propagate an error down the observable?

flotothemoon, line 35: Considering these BFT checks are all just assertions, it might be more apt to call them that (BFTAssertions).

flotothemoon: That is, unless we have any intention of supporting non-assertive checks?

radixdlt/src/integration/java/com/radixdlt/consensus/checks/LivenessCheck.java
```diff
@@ -0,0 +1,64 @@
...
+ * Check that the network is making progress by ensuring that QCs are progressively
+ * increasing.
+ */
+public class LivenessCheck implements BFTCheck {
+
+	@Override
+	public Observable<Object> check(BFTTestNetwork network) {
+		// there should be a new highest QC every once in a while to ensure progress
+		// the minimum latency per round is determined using the network latency
+		// a round can consist of 6 * MTT
```
flotothemoon: MTT meaning?

radixdlt/src/integration/java/com/radixdlt/consensus/checks/NoSyncExceptionCheck.java
```diff
@@ -0,0 +1,42 @@
...
+import org.assertj.core.api.Condition;
+
+/**
+ * Checks that there are no synchronisation errors.
+ */
+public class NoSyncExceptionCheck implements BFTCheck {
+
+	@Override
+	public Observable<Object> check(BFTTestNetwork network) {
+		return Observable.interval(1, TimeUnit.SECONDS)
```
flotothemoon, line 35: I think the time-based periodic checks would benefit from having the time periods injected as parameters (e.g. for testing)

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
```diff
@@ -115,37 +111,40 @@ public void setReceivingDisable(ECPublicKey validatorId, boolean disable) {
...
+		final Consumer<MessageInTransit> sendMessageSink = message -> {
+			if (sendingDisabled.contains(forNode)) {
+				return;
+			}
+
+			final int nextLatency;
+			if (message.target.equals(forNode)) {
+				nextLatency = 0;
+			} else {
+
```
flotothemoon: Unnecessary newline

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
```diff
@@ -115,37 +111,40 @@ public void setReceivingDisable(ECPublicKey validatorId, boolean disable) {
...
+				final long curTime = System.currentTimeMillis();
+				final Long lastTimestamp = lastTimestamps.get(message.target);
```
flotothemoon: Ha, funny enough I considered doing it this way as well before deciding on the queue-based approach. My reasoning was that it would be harder to ensure preserved ordering in a threadsafe way without the queue, with the added limitation that there is no easy way to view the currently pending messages (which I've found useful for debugging). 

Specifically, in my limited understanding, I'm not confident the current implementation is thread-safe because the code between reading the timestamp (line 128) and writing the next timestamp back (line 136) is not synchronized.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -69,11 +69,28 @@ public VertexStore(
...
+				// TODO: remove blocking
+				proposedVertex = vertexSupplier.getVertex(qc.getProposed().getId()).blockingGet();
+			} catch (Exception e) {
+				throw new SyncException(qc, e);
+			}
+
+			try {
+				this.insertVertex(proposedVertex);
+			} catch (VertexInsertionException e) {
+				throw new IllegalStateException("Should not go here.");
```
flotothemoon: Should not go here because of the earlier check (line 75)? Maybe add a comment to that effect and include the exception as a cause to the `IllegalStateException`, the trigger for `VertexInsertionException` may change in the future

radixdlt/src/integration/java/com/radixdlt/consensus/checks/LivenessCheck.java
```diff
@@ -0,0 +1,64 @@
...
+		double trips = 6.0;
+		int maxLatencyPerRound = (int) (network.getMaximumNetworkLatency() * trips);
+		int maxNumRoundsBeforeProgress = 2;
```
flotothemoon: The number of maximum rounds before progress depends on a few factors, like:
- The exact leader rotation algorithm
- The number of potential leaders 

So `maxNumRoundsBeforeProgress` can vary quite a bit, but exposing it as a parameter directly is a bit odd as well. Maybe just inject it as a constant for now?

radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
```diff
@@ -0,0 +1,113 @@
...
+	private final TimeUnit timeUnit;
+	private final int minNetworkLatency;
+	private final int maxNetworkLatency;
+	private final List<BFTCheck> checks = new ArrayList<>();
```
flotothemoon, line 40: I'm curious what your plans for dynamic configurations are, which may require different checks over time. I had a dynamic configuration in an earlier test where nodes randomly die over time, and this sort of dynamic is useful to test for. Will we be mapping configurations to sets of checks?

radixdlt/src/integration/java/com/radixdlt/consensus/checks/AllProposalsHaveDirectParentsCheck.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public class AllProposalsHaveDirectParentsCheck implements BFTCheck {
+
+	@Override
+	public Observable<Object> check(BFTTestNetwork network) {
+		List<Observable<Vertex>> correctProposals = network.getNodes().stream()
+			.map(ECKeyPair::getPublicKey)
+			.map(network.getUnderlyingNetwork()::getNetworkRx)
+			.map(EventCoordinatorNetworkRx::consensusEvents)
+			.map(p -> p.ofType(Proposal.class).map(Proposal::getVertex))
```
flotothemoon, line 44: Is the second .map intended to be on the same line? 
Also, I think Observable supports ofType natively, would save all of 3 characters :)

radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+	private BFTTest(int numNodes, long time, TimeUnit timeUnit, int minNetworkLatency, int maxNetworkLatency) {
+		this.numNodes = numNodes;
+		this.time = time;
+		this.timeUnit = timeUnit;
+		this.minNetworkLatency = minNetworkLatency;
+		this.maxNetworkLatency = maxNetworkLatency;
+	}
+
+	public static class BFTTestBuilder {
```
clstrfsck: Should really be a `public static BFTTestBuilder builder()` method, so that we can do things like `BFTTest foo = BFTTest.builder().this().that().other().build()`, rather than relying on constructors to build.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
clstrfsck: Agree with @flotothemoon , needs comment and chained exception

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -69,11 +69,28 @@ public VertexStore(
...
+				// Too far behind
+				// TODO: more syncing
+				throw new SyncException(qc);
+			}
+
+			final Vertex proposedVertex;
+			try {
+				// TODO: remove blocking
+				proposedVertex = vertexSupplier.getVertex(qc.getProposed().getId()).blockingGet();
+			} catch (Exception e) {
```
clstrfsck, line 85: Can we catch the actual valid exceptions here, rather than just everything?

radixdlt/src/main/java/com/radixdlt/consensus/View.java
```diff
@@ -58,7 +58,7 @@ public boolean equals(Object o) {
 
 	@Override
 	public String toString() {
-		return "View " + this.view;
+		return Long.toString(this.view);
```
clstrfsck, line 78: How does this class *still* have no header comment?

Also, String.valueOf is more common than Long.toString, but doesn't really matter.

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
```diff
@@ -28,13 +28,12 @@
 import io.reactivex.rxjava3.subjects.PublishSubject;
 
 import java.util.Collections;
-import java.util.Deque;
```
clstrfsck, line 31: This class looks like it is in the wrong part of the source tree.  Should be in radixdlt/src/test/java

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
```diff
@@ -46,16 +45,14 @@
 	private final Random rng;
 	private final int minimumLatency;
 	private final int maximumLatency;
-	private final boolean preserveOrder;
 
-	private final Deque<MessageInTransit> orderedMessageBuffer;
 	private final PublishSubject<MessageInTransit> receivedMessages;
 	private final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor();
+	private final Set<ECPublicKey> readers = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
clstrfsck: Could use `Sets.newConcurrentHashSet()`

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
```diff
@@ -115,37 +111,40 @@ public void setReceivingDisable(ECPublicKey validatorId, boolean disable) {
...
-	}
-
 	public EventCoordinatorNetworkSender getNetworkSender(ECPublicKey forNode) {
-		Consumer<MessageInTransit> sendMessageSink = message -> {
-			if (!sendingDisabled.contains(forNode)) {
-				if (preserveOrder) {
-					orderedMessageBuffer.push(message);
+		final Map<ECPublicKey, Long> lastTimestamps = new ConcurrentHashMap<>();
+
+		final Consumer<MessageInTransit> sendMessageSink = message -> {
```
clstrfsck: This seems pretty ordinary.  It looks like this is here just so we can include this method inside another method.

Really should be broken out into a separate method.

clstrfsck (changes requested):
Looks like there are some structural things that need attention, esp @flotothemoon comment about thread safety.

radixdlt/src/integration/java/com/radixdlt/consensus/checks/LivenessCheck.java
talekhinezh: max transmission time

radixdlt/src/integration/java/com/radixdlt/consensus/checks/LivenessCheck.java
talekhinezh: true

radixdlt/src/integration/java/com/radixdlt/consensus/checks/NoSyncExceptionCheck.java
talekhinezh: true but not gonna add it until we actually need it

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
talekhinezh: I think this module is hefty enough and gonna need its own unit tests so `/src/test/java` doesn't seem appropriate. Thoughts?

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: Keeping it to Exception for now as an actual implementation to `getVertex` hasn't been implemented yet but will add comment.

talekhinezh: unfortunately not, since we are using blocking method of RX, we can't catch CheckedExceptions

radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
talekhinezh: haven't really gotten this far yet into thinking how that would look like

flotothemoon: Ah, so `CrashFaultNetworkTest` are staying as they are for now and not integrated with this new BFTTest interface (specifically, `given_7_correct_bfts_that_randomly_crash__then_correct_instances_should_make_progress_as_possible_over_1_minute`).

radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
talekhinezh: yeah will do that later methinks

radixdlt/src/integration/java/com/radixdlt/consensus/BFTCheck.java
talekhinezh: Wanted to stay away from well known `assert` because the way we need to check is not the traditional `assertX` method.

radixdlt/src/integration/java/com/radixdlt/consensus/BFTTest.java
talekhinezh: yup

radixdlt/src/integration/java/com/radixdlt/consensus/checks/AllProposalsHaveDirectParentsCheck.java
talekhinezh: yup intended. not sure what you're suggesting here? one is a stream the other an observable

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
clstrfsck: There's no problem with writing unit tests for unit tests :)

I think from a deliverable point of view, it would be better to exclude code from the runtime package that wasn't required for runtime.  Some important security benefits, including smaller code footprint, a no real drawbacks that I can see.

radixdlt/src/integration/java/com/radixdlt/consensus/checks/AllProposalsHaveDirectParentsCheck.java
flotothemoon: Ah man, I totally misread this line. Nevermind.

radixdlt/src/integration/java/com/radixdlt/consensus/checks/LivenessCheck.java
flotothemoon: Please add a comment stating that, wasn't obvious to me

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
```diff
@@ -115,87 +104,80 @@ public void setReceivingDisable(ECPublicKey validatorId, boolean disable) {
...
+			.filter(msg -> !receivingDisabled.contains(msg.target))
+			.filter(msg -> msg.target.equals(forNode))
+			.timestamp(TimeUnit.MILLISECONDS)
+			.scan((msg1, msg2) -> {
+				if (msg2.value().sender.equals(forNode)) {
+					return msg2;
+				}
+				int delayCarryover = (int) Math.max(msg1.time() + msg1.value().delay - msg2.time(), 0);
+				int range = maximumLatency - delayCarryover - minimumLatency + 1;
+				// Currently a gaussian distribution
```
clstrfsck: Why gaussian?  Looks uniform to me.

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
```diff
@@ -115,87 +104,80 @@ public void setReceivingDisable(ECPublicKey validatorId, boolean disable) {
...
+				int nextDelay = range > 0 ? minimumLatency + rng.nextInt(range) : 0;
```
clstrfsck: Should this be `range > 0 ? minimumLatency + rng.nextInt(range) : Math.max(0, minimumLatency + range)`?

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
flotothemoon: This has been marked resolved, yet the code wasn't changed here - can you clarify @talekhinezh?

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
clstrfsck: Or maybe `range > 0 ? minimumLatency + rng.nextInt(range) : minimumLatency`?

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
talekhinezh: sorry yeah, mixed up

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: missed "cause" comment

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
talekhinezh: I think your first comment is right

talekhinezh: or rather it should just be: `range > 0 ? minimumLatency + rng.nextInt(range) : minimumLatency + range`

radixdlt/src/main/java/com/radixdlt/middleware2/network/TestEventCoordinatorNetwork.java
talekhinezh: I think these simulation tests need some re-thinking on where they belong so I'll just add another task for that.

clstrfsck (approved):

PR #74, opened 2020-04-22
Bugfix/rpnv1 167 linear quorum

clstrfsck:
Changes to ensure that quorums are formed with a linear number of signature verifications.


radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidationState.java
```diff
@@ -0,0 +1,128 @@
...
+import com.radixdlt.crypto.ECDSASignature;
+import com.radixdlt.crypto.ECDSASignatures;
+import com.radixdlt.crypto.ECPublicKey;
+import com.radixdlt.crypto.Hash;
+
+/**
+ * Keeps track of current validation state for a thing that
+ * needs multiple correct signatures for a quorum.
+ */
+public final class ValidationState {
```
flotothemoon, line 35: It feels a bit odd to conflate the requirements for a quorum (`acceptableFaults` and all that) with the state used to keep track of signatures, checking against those requirements. Probably fine for now, just wanted to flag.

flotothemoon (approved):
Looking splendid. 
Would be great to have a test that demos the performance gains we would like to see here.

clstrfsck:
> Looking splendid.
> Would be great to have a test that demos the performance gains we would like to see here.

The `ValidationStateTest. testLinearSignatureValidation()` test ensures that there are is at most one verify operation for each signature when forming a quorum.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidationState.java
clstrfsck: Could rename to QuorumState, I guess.  Not sure it really makes sense to break the thresholding out into something separate right now, although I feel kind of sure that this is going to be subject to some change in the future in any case.

PR #77, opened 2020-04-23
Initial implementation of test data counters and APIs

clstrfsck:


radixdlt/src/main/java/org/radix/api/http/RadixHttpServer.java
```diff
@@ -105,210 +122,220 @@ public RadixHttpServer(
...
+			.subscribe(this::vertexUpdate);
+	}
 
-    /**
-     * Get the set of currently connected peers
-     *
-     * @return The currently connected peers
-     */
-    public final Set<RadixJsonRpcPeer> getPeers() {
-        return Collections.unmodifiableSet(peers.keySet());
```
clstrfsck, line 121: Lots of spaces converted to tabs in this file - might be easier to review with whitespace off.

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
```diff
@@ -217,17 +218,21 @@ public void processProposal(Proposal proposal) {
...
 
 		final View updatedView = this.pacemaker.getCurrentView();
-		if (proposedVertex.getView().compareTo(updatedView) != 0) {
-			log.info("{}: PROPOSAL: Ignoring view {} Current is: {}", this.getShortName(), proposedVertex.getView(), updatedView);
+		if (proposedVertexView.compareTo(updatedView) != 0) {
+			log.info("{}: PROPOSAL: Ignoring view {} Current is: {}", this.getShortName(), proposedVertexView, updatedView);
 			return;
 		}
 
+		if (proposedVertex.getParentView() != null && proposedVertexView.number() != proposedVertex.getParentView().number() + 1) {
```
talekhinezh: proposedVertex.getParentView() should never be null. This should be a validation/assertion check instead if it is to be used.

Also, Vertex has `.hasDirectParent()` which can be used instead

talekhinezh (approved):

PR #79, opened 2020-04-27
Implement voting power (Review)

talekhinezh:
* Implement weighted votes based on the power of a validator
* Create a new ProposerElection called WeightedRotatingLeaders which rotates leaders based on power of each validator


radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidationState.java
```diff
@@ -77,7 +77,8 @@ public boolean addSignature(ECPublicKey key, ECDSASignature signature) {
 	 * @return {@code true} if we have enough valid signatures to form a quorum,
 	 */
 	public boolean complete() {
-		return this.signedKeys.size() >= threshold(this.validKeys.size());
+		UInt256 currentPower = this.validatorSet.getPower(this.signedKeys.keySet());
```
clstrfsck: Unfortunately this is no longer O(1).  Probably need to keep track of each validators power as the signatures are added.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,160 @@
...
+			this.validatorSet = validatorSet;
+			this.weightsComparator = weightsComparator;
+			this.weights = new HashMap<>();
+			this.cache = new Validator[cacheSize];
+			this.powerArray = validatorSet.getValidators().stream().map(Validator::getPower).toArray(UInt128[]::new);
+			this.resetToView(View.of(0));
+		}
+
+		private Validator computeHeaviest() {
+			final Entry<Validator, UInt256> max = weights.entrySet().stream()
```
clstrfsck, line 77: This is constant for a given validator set, so should be precomputed.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		private Validator resetToView(View view) {
+			// reset if view isn't in cache
+			if (curView == null || view.number() < curView.number() - cache.length) {
+				// after cappedLCM is executed, the following invariant will be true:
+				// (lcm > 0 && lcm < 2^64) || lcm == null
+				// This is due to use of 2^64 cap and also the invariant from ValidatorSet
+				// that powerArray will always be non-zero
+				UInt128 lcm = MathUtils.cappedLCM(UInt128.from(view.number()), powerArray);
```
clstrfsck: LCM is constant for validator set, so should be precomputed.  Can be capped at Long.MAX_VALUE.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,160 @@
...
+				if (lcm == null) {
+					curView = View.of(0);
```
clstrfsck: View.genesis() ?

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidationState.java
```diff
@@ -77,7 +77,8 @@ public boolean addSignature(ECPublicKey key, ECDSASignature signature) {
...
+		return currentPower.compareTo(threshold(this.validatorSet.getTotalPower())) >= 0;
```
clstrfsck: Should precompute threshold, as this is relatively expensive, but fixed for given validator set.

radixdlt/src/main/java/com/radixdlt/consensus/validators/Validator.java
```diff
@@ -85,46 +64,35 @@
 	private final boolean jailed;
 */
 
-    // Public key for consensus
-	private ECPublicKey nodeKey;
+	// Power associated with each validator, could e.g. be based on staked tokens
+	private final UInt128 power;
```
clstrfsck: As discussed, I think this is going to be very much easier in future if we use a UInt256 here.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -57,7 +70,24 @@ public static ValidatorSet from(Collection<Validator> validators) {
...
 	public ValidationState newValidationState(Hash anchor) {
-		return ValidationState.forValidatorSet(anchor, validators.keySet());
+		return ValidationState.forValidatorSet(anchor, this);
+	}
+
+	public boolean containsKey(ECPublicKey key) {
+		return validators.containsKey(key);
+	}
+
+	public UInt256 getPower(Set<ECPublicKey> signedKeys) {
```
clstrfsck: See other comments about how this makes complete() method non-constant time.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -57,7 +70,24 @@ public static ValidatorSet from(Collection<Validator> validators) {
 	 * @return An initial validation state with no signatures
 	 */
 	public ValidationState newValidationState(Hash anchor) {
-		return ValidationState.forValidatorSet(anchor, validators.keySet());
+		return ValidationState.forValidatorSet(anchor, this);
+	}
+
+	public boolean containsKey(ECPublicKey key) {
+		return validators.containsKey(key);
```
clstrfsck, line 79: Two methods the same name, but with different meaning of "key" amused me here.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -57,7 +70,24 @@ public static ValidatorSet from(Collection<Validator> validators) {
...
+	}
+
+	public UInt256 getPower(Set<ECPublicKey> signedKeys) {
+		return signedKeys.stream()
+			.map(validators::get)
+			.map(Validator::getPower)
```
clstrfsck: Should filter nulls after validators::get, as there is no enforcing that all the signedKeys are required to be in the validator set from the interface.

This perhaps could be specified in the Javadoc for this method, but there is none.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,160 @@
...
+ *
+ * This class stateful and is NOT thread-safe.
+ */
+public final class WeightedRotatingLeaders implements ProposerElection {
+
+	private final ValidatorSet validatorSet;
+	private final Comparator<Entry<Validator, UInt256>> weightsComparator;
+	private final CachingNextLeaderComputer nextLeaderComputer;
+
+	public WeightedRotatingLeaders(ValidatorSet validatorSet, Comparator<Validator> comparator, int cacheSize) {
```
clstrfsck, line 53: Would be better IMO to document that tie breaks are handled through iterator ordering of the underlying container, rather than passing in a comparator to handle them.

That way, callers can control tie break resolution without complicating the interface and operation here.

clstrfsck (changes requested):
We have made quorum checking non-linear again, and this needs to be fixed.

Other minor suggestions and improvements.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
talekhinezh: lol

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
talekhinezh: I would agree but only issue is that `ValidatorSet` has no concept of a deterministic order of validators. Perhaps it should but if the case I think that should be reserved for another task/PR.

clstrfsck: I think this is the case as things stand as ValidatorSet is based on an ImmutableBiMap, which retains the deterministic iteration on insertion order per the ImmutableCollection contract.

https://guava.dev/releases/29.0-jre/api/docs/com/google/common/collect/ImmutableCollection.html

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,174 @@
...
+			if (curView == null || view.number() < curView.number() - cache.length) {
+				if (lcm == null || lcm > view.number()) {
+					curView = View.genesis();
+				} else {
+					long multipleOfLCM = view.number() / lcm;
+					curView = View.of(multipleOfLCM * lcm);
+				}
+
+				for (Validator validator : validatorSet.getValidators()) {
+					weights.put(validator, UInt384.from(UInt128.ONE, UInt256.ZERO).subtract(validator.getPower()));
```
clstrfsck: Would be better as a constant, rather than creating a new one only to throw away immediately.
```
private static final UInt384 POW_2_256 = UInt384.from(UInt256.MAX_VALUE).increment();
```

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -27,21 +28,35 @@
...
+ * Set of validators for consensus. Only validators with power >= 1 will
+ * be part of the set.
  * <p>
  * Note that this set will validate for set sizes less than 4,
  * as long as all validators sign.
  */
 public final class ValidatorSet {
 	private final ImmutableBiMap<ECPublicKey, Validator> validators;
 
+	// We assume that we won't have more than 2^128 validators (2^32 is in fact the limit)
```
clstrfsck: I think this comment is not quite right anymore.  We assume that there will be less than 2^256 staking tokens.

radixdlt/build.gradle
```diff
@@ -83,7 +83,10 @@ jacocoTestReport {
 }
 
 dependencies {
-    compile 'com.radixdlt:radix-engine-library:1.0-beta.8'
+    compile 'com.radixdlt:radixdlt-java-common:63d1dec1c7'
+    compile ('com.radixdlt:radix-engine-library:1.0-beta.8') {
+        exclude group: 'com.radixdlt', module: 'radixdlt-java-common'
```
clstrfsck, line 88: Ewww.  But I guess it works.

talekhinezh: will get resolved come release time

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
clstrfsck: Per discussion, deferring any changes here until interfaces settle down a bit.

clstrfsck (approved):
Looks good to me.

PR #82, opened 2020-04-27
Implement voting power (Review Only)

talekhinezh:


radixdlt/src/main/java/com/radixdlt/consensus/validators/Validator.java
```diff
@@ -85,27 +80,40 @@
...
 	}
 
+
 	private Validator(
-		ECPublicKey nodeKey
+		ECPublicKey nodeKey,
+		UInt256 power
 	) {
 		this.nodeKey = Objects.requireNonNull(nodeKey);
+		this.power = Objects.requireNonNull(power);
```
flotothemoon, line 78: Does it make sense for a node to have zero power then?

radixdlt/src/test/java/com/radixdlt/consensus/liveness/WeightedRotatingLeadersTest.java
```diff
@@ -0,0 +1,140 @@
...
+				}
+				ECPublicKey pk1 = weightedRotatingLeaders.getProposer(View.of(viewsToTest - 1));
+				ECPublicKey pk2 = weightedRotatingLeaders2.getProposer(View.of(viewsToTest - 1));
+				assertThat(pk1).isEqualTo(pk2);
+			}
+		}
+	}
+
+	@Test
+	public void when_validators_distributed_by_fibonacci__then_leaders_also_distributed_in_fibonacci() {
```
flotothemoon, line 112: What a beautiful test.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -33,11 +36,23 @@
  * as long as all validators sign.
  */
 public final class ValidatorSet {
+	// We assume that we won't have more than 2^128 validators (2^32 is in fact the limit)
+	// in a single validator set so having this as the max power value will prevent overflows
+	private static final UInt256 POWER_MAX_VALUE = UInt256.from(UInt128.MAX_VALUE);
 	private final ImmutableBiMap<ECPublicKey, Validator> validators;
+	private final transient UInt256 totalPower;
```
flotothemoon, line 43: Why is this marked transient, seeing as its tied to `validators` and not memoized?

radixdlt/src/main/java/com/radixdlt/consensus/validators/Validator.java
```diff
@@ -85,27 +80,40 @@
 	private final boolean jailed;
 */
 
+	@JsonProperty("power")
+	@DsonOutput(Output.ALL)
+	// Staked tokens
```
flotothemoon: Does `power` necessarily equate "staked tokens" as per the comment?

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,142 @@
...
+				for (Validator validator : validatorSet.getValidators()) {
+					weights.put(validator, UInt256.from(UInt128.ONE, UInt128.ZERO).subtract(validator.getPower()));
+				}
+				cache[0] = computeHeaviest();
+			}
+
+			// compute to view
+			computeToView(view);
+
+			// guaranteed to return non-null;
```
flotothemoon, line 139: Could annotate method with `@NonNull` or similar then

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,142 @@
...
+				computeNext();
+			}
+		}
+
+		private Validator resetToView(View view) {
+			// reset if view isn't in cache
+			if (curView == null || view.number() < curView.number() - cache.length) {
+				curView = View.of(0);
+				for (Validator validator : validatorSet.getValidators()) {
+					weights.put(validator, UInt256.from(UInt128.ONE, UInt128.ZERO).subtract(validator.getPower()));
```
flotothemoon, line 131: Is the subtraction here supposed to underflow? Or are we assuming validator.getPower() to be <2^128?

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,142 @@
...
+
+			// guaranteed to return non-null;
+			return cache[(int) (view.number() % cache.length)];
+		}
+	}
+
+	@Override
+	public ECPublicKey getProposer(View view) {
+		nextLeaderComputer.computeToView(view);
+		Validator validator = nextLeaderComputer.checkCacheForProposer(view);
```
flotothemoon, line 150: The only case I can see where this cache lookup will fail is if `view` is before `nextLeaderComputer`'s cache window (since in the other case, the preceding call to `computeToView` will catch up to `view`).
If my understanding is correct, it would be good to describe that here.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,142 @@
...
+				.max(weightsComparator)
+				.orElseThrow(() -> new IllegalStateException("Weights cannot be empty"));
+			return max.getKey();
+		}
+
+		private void computeNext() {
+			// Reset last leader
+			final int curIndex = (int) (this.curView.number() % cache.length);
+			final Validator curLeader = cache[curIndex];
+			weights.merge(curLeader, validatorSet.getTotalPower(), UInt256::subtract);
```
flotothemoon, line 87: Took me a bit to get why this works. Please add a few comments to the lines where weights are modified to explain the idea (reduce by total and then accumulate with power so that the leader will rise back to the top with a pace corresponding to their relative power)

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
flotothemoon: (if so, need to test that assumption here, or maybe even change the data type of `Validator.power`)

flotothemoon (changes requested):
Generally good. I would like to see a few things clarified that confused me a bit.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
talekhinezh: good point

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
talekhinezh: assuming validator.getPower() < 2^128. This is part of ValidatorSet contract but gonna change this so that power is just 128 bits so it becomes compile time check

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
talekhinezh: yup thats right

radixdlt/src/main/java/com/radixdlt/consensus/validators/Validator.java
talekhinezh: Doesn't break the algorithm in any way

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
talekhinezh: It's basically a cache of information already in validators. Also affects EqualsVerifier.

radixdlt/src/test/java/com/radixdlt/consensus/liveness/WeightedRotatingLeadersTest.java
talekhinezh: :)

radixdlt/src/main/java/com/radixdlt/consensus/validators/Validator.java
```diff
@@ -85,46 +64,35 @@
...
 
-    // Public key for consensus
-	private ECPublicKey nodeKey;
+	// Power associated with each validator, could e.g. be based on staked tokens
+	private final UInt128 power;
 
-	Validator() {
-		// Serializer only
-	}
+    // Public key for consensus
```
flotothemoon, line 70: Some odd formatting here.

flotothemoon (approved):
Good to go now.

PR #85, opened 2020-04-28
Refactor pacemaker (Review)

talekhinezh:
* Separate out pacemaker state machine from timeout events emitted


radixdlt/src/main/java/com/radixdlt/consensus/liveness/FixedTimeout.java
```diff
@@ -19,68 +19,53 @@
...
 import java.util.Optional;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 
 /**
- * Overly simplistic pacemaker
+ * A pacemaker which utilizes a fixed timeout (aka requires a synchronous network).
  */
-public final class PacemakerImpl implements Pacemaker, PacemakerRx {
+public final class FixedTimeout implements Pacemaker {
```
clstrfsck: Should really be FixedTimeoutPacemaker.  At the moment reads like this is a class implementing a timeout rather than a pacemaker.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/MempoolProposalGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+	// TODO: check that next proposal works with current vertexStore state
+	@Override
+	public Vertex generateProposal(View view) {
+		final QuorumCertificate highestQC = vertexStore.getHighestQC();
+		final List<Vertex> preparedVertices = vertexStore.getPathFromRoot(highestQC.getProposed().getId());
+		final Set<AID> preparedAtoms = preparedVertices.stream()
+			.filter(v -> v.getAtom() != null)
+			.map(Vertex::getAtom)
+			.map(Atom::getAID)
+			.collect(Collectors.toSet());
```
clstrfsck, line 58: Would be slightly better as:
```
		final Set<AID> preparedAtoms = preparedVertices.stream()
			.map(Vertex::getAtom)
			.filter(Objects::nonNull)
			.map(Atom::getAID)
			.collect(Collectors.toSet());
```

radixdlt/src/main/java/com/radixdlt/consensus/liveness/MempoolProposalGenerator.java
```diff
@@ -0,0 +1,47 @@
+package com.radixdlt.consensus.liveness;
```
clstrfsck, line 18: Missing header comments.

radixdlt/src/test/java/com/radixdlt/consensus/liveness/FixedTimeoutTest.java
```diff
@@ -20,118 +20,72 @@
...
-	@Test
-	public void when_start__then_a_timeout_event_with_view_0_is_emitted() {
-		ScheduledExecutorService executorService = getMockedExecutorService();
-		PacemakerImpl pacemaker = new PacemakerImpl(TEST_PACEMAKER_TIMEOUT, executorService);
-		TestObserver<View> testObserver = TestObserver.create();
-		pacemaker.localTimeouts().subscribe(testObserver);
-		testObserver.awaitCount(1);
-		testObserver.assertNotComplete();
-		testObserver.assertValues(View.of(0L));
+		assertThatThrownBy(() -> new FixedTimeout(0, mock(FixedTimeout.TimeoutSender.class)));
```
clstrfsck: Really should be checking the type of the exception.

radixdlt/src/test/java/com/radixdlt/consensus/liveness/ScheduledTimeoutSenderTest.java
```diff
@@ -0,0 +1,104 @@
...
+		testObserver.awaitCount(2);
+		testObserver.assertNotComplete();
+		testObserver.assertValues(view1, view2);
+		verify(executorService, times(2)).schedule(any(Runnable.class), eq(timeout), eq(TimeUnit.MILLISECONDS));
+	}
+
+	@Test
+	public void when_subscribed_to_timeout_and_timeout_occurs_for_view__then_should_complete() throws Exception {
+		View view = View.of(0);
+		TestObserver testObserver = TestObserver.create();
```
clstrfsck: Missing type argument

radixdlt/src/test/java/com/radixdlt/consensus/liveness/ScheduledTimeoutSenderTest.java
```diff
@@ -0,0 +1,104 @@
...
+		testObserver.await();
+		testObserver.assertComplete();
+	}
+
+	@Test
+	public void when_subscribe_to_timeout_and_timeout_already_occurred__then_should_complete() throws Exception {
+		View view = View.of(0);
+		scheduledTimeoutSender.scheduleTimeout(view, 10);
+		scheduledTimeoutSender.scheduleTimeout(view.next(), 10);
+		TestObserver testObserver = TestObserver.create();
```
clstrfsck: Missing type argument

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTNetworkSimulation.java
```diff
@@ -177,7 +185,7 @@ public SystemCounters getCounters(ECKeyPair keyPair) {
 		return counters.get(keyPair);
 	}
 
-	public PacemakerImpl getPacemaker(ECKeyPair keyPair) {
+	public FixedTimeout getPacemaker(ECKeyPair keyPair) {
```
clstrfsck: Return type probably should be Pacemaker rather than FixedTimeout

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
```diff
@@ -113,7 +117,7 @@ private void sync(QuorumCertificate qc, ECPublicKey node) throws SyncException {
 					this.getShortName(), this.getShortName(node.euid()), vertexId.toString().substring(0, 6));
 				return networkSender.getVertex(vertexId, node)
 					.doOnSuccess(v -> log.info("{}: Received GET_VERTEX Response: {}", this.getShortName(), v))
-					.takeUntil(this.pacemaker.nextLocalTimeout());
+					.takeUntil(this.pacemakerRx.timeout(this.pacemaker.getCurrentView()));
```
flotothemoon, line 120: Is there a possible race condition between the call to `timeout` and `getCurrentView` here?

flotothemoon (approved):
Looks good to me.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,142 @@
...
+
+	public WeightedRotatingLeaders(ValidatorSet validatorSet, Comparator<Validator> comparator, int cacheSize) {
+		this.validatorSet = validatorSet;
+		this.weightsComparator = Comparator
+			.comparing(Entry<Validator, UInt256>::getValue)
+			.thenComparing(Entry::getKey, comparator);
+		this.nextLeaderComputer = new CachingNextLeaderComputer(validatorSet, weightsComparator, cacheSize);
+	}
+
+	private static class CachingNextLeaderComputer {
```
clstrfsck: This looks like weighted-round-robin scheduling see here: http://kb.linuxvirtualserver.org/wiki/Weighted_Round-Robin_Scheduling

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,142 @@
...
+			return null;
+		}
+
+		private void computeToView(View view) {
+			while (view.compareTo(curView) > 0) {
+				computeNext();
+			}
+		}
+
+		private Validator resetToView(View view) {
```
clstrfsck: It's not necessary to reset to 0.  Can reset to largest multiple of Least-Common-Multiple(validatorSet powers) that is less than the view number, product a UInt512 etc.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,142 @@
...
+			// reset if view isn't in cache
+			if (curView == null || view.number() < curView.number() - cache.length) {
+				curView = View.of(0);
+				for (Validator validator : validatorSet.getValidators()) {
+					weights.put(validator, UInt256.from(UInt128.ONE, UInt128.ZERO).subtract(validator.getPower()));
```
clstrfsck: Generally this should be done better.  If we are making validator power a UInt256, then the sum of all validator powers should be a UInt384 (for example).

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -36,11 +36,23 @@
  * as long as all validators sign.
  */
 public final class ValidatorSet {
+	// We assume that we won't have more than 2^128 validators (2^32 is in fact the limit)
+	// in a single validator set so having this as the max power value will prevent overflows
+	private static final UInt256 POWER_MAX_VALUE = UInt256.from(UInt128.MAX_VALUE);
 	private final ImmutableBiMap<ECPublicKey, Validator> validators;
+	private final transient UInt256 totalPower;
```
clstrfsck: I don't think we need to worry about "transient" modifier unless there is some likelihood that the object will be serialized in some way.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -36,11 +36,23 @@
  * as long as all validators sign.
  */
 public final class ValidatorSet {
+	// We assume that we won't have more than 2^128 validators (2^32 is in fact the limit)
+	// in a single validator set so having this as the max power value will prevent overflows
+	private static final UInt256 POWER_MAX_VALUE = UInt256.from(UInt128.MAX_VALUE);
```
clstrfsck: See other comments on validator power.  Should either make this a UInt128, or a UInt256, but either way allow the full range of the underlying type.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
clstrfsck: Probably should be UInt384.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
```diff
@@ -54,39 +66,33 @@ public static ValidatorSet from(Collection<Validator> validators) {
...
+		return validators.containsKey(key);
 	}
 
-	@VisibleForTesting
-	static int threshold(int n) {
-		return n - acceptableFaults(n);
+	public UInt256 getPower(Set<ECPublicKey> signedKeys) {
+		return signedKeys.stream()
+			.map(validators::get)
+			.map(Validator::getPower)
```
clstrfsck: Should filter out null values here?

radixdlt/src/main/java/com/radixdlt/consensus/ValidatingEventCoordinator.java
talekhinezh: nope, `.timeout` works whether the timeout already occurred or not via use of `BehaviorSubject` in ScheduledTimeoutSender. Rx magic!

clstrfsck (changes requested):
Need to fix sizing for power calculations (UInt128/UInt256/UInt384/UInt512), as this seems a bit incoherent and fragile at the moment.  At least, the sum of numerous UInt256 values should be UInt384.

I think there are some improvements that can be made in the round-robin scheduler for leader selection.  References to the Linux WRR scheduler have been provided.  In addition, there is a simple improvement that could be made by starting at the largest multiple of the least-common-multiple that is less than the target view, rather than current approach of starting from zero each time.  It takes about 200 seconds to handle 10,000,000 views on my machine this way, and that is likely only 3-4 days of network runtime.  After a year, it might take a node 5-6 hours to start.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/FixedTimeout.java
talekhinezh: true

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
talekhinezh: yup this is resolved in https://github.com/radixdlt/radixdlt-core/pull/82 which this PR is dependent on

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
talekhinezh: nice!

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
talekhinezh: yup resolved in https://github.com/radixdlt/radixdlt-core/pull/82

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
talekhinezh: `transient` is also used by EqualsVerifier to see what should actually be checked for equality

clstrfsck: I would prefer that we placate EqualsVerifier in the setup for EqualsVerifier, rather than in the code proper.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * This class stateful and is NOT thread-safe.
+ */
+public final class WeightedRotatingLeaders implements ProposerElection {
+
+	private final ValidatorSet validatorSet;
+	private final Comparator<Entry<Validator, UInt256>> weightsComparator;
+	private final CachingNextLeaderComputer nextLeaderComputer;
+
+	public WeightedRotatingLeaders(ValidatorSet validatorSet, Comparator<Validator> comparator, int cacheSize) {
```
clstrfsck: Not sure what the purpose of the comparator is here, given that the only real option is the validator power, and the power already is comparable?

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
talekhinezh: Seems like this algorithm though uses less state is around O(n^2 / gcd(S)) in compute, no? Also, seems like their algorithm is less able to handle dynamic servers coming in and out

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
talekhinezh: It's possible for power and weights to be equal and need a deterministic way of choosing the winner

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
talekhinezh: Not sure I agree here. I think `transient` is clear indicator in code that parameter does not participate in equals and is most likely derived from other parts of the classes state. Also, a clear indicator if this class is to be serialized by some serialization mechanism in the future.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
talekhinezh: every validator object is required to have power even if zero

clstrfsck: Moved to #79

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
clstrfsck: Moved to #79

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
```diff
@@ -72,10 +76,12 @@ public CerberusModule(RuntimeProperties runtimeProperties) {
 	protected void configure() {
 		// dependencies
 		bind(Scheduler.class).toProvider(SingleThreadedScheduler::new);
-		bind(PacemakerRx.class).to(PacemakerImpl.class);
-		bind(Pacemaker.class).to(PacemakerImpl.class);
+		bind(TimeoutSender.class).to(ScheduledTimeoutSender.class);
+		bind(PacemakerRx.class).to(ScheduledTimeoutSender.class);
+		bind(Pacemaker.class).to(FixedTimeoutPacemaker.class);
```
clstrfsck: Configuration here seems to have been a bit messed up over time.  This binding should be removed.

See also FixedTimeoutPacemaker provider below, which should be a Pacemaker provider.

radixdlt/src/main/java/com/radixdlt/CerberusModule.java
```diff
@@ -100,14 +106,22 @@ private EpochRx epochRx(
...
 	@Provides
 	@Singleton
-	private PacemakerImpl pacemaker() {
+	private ScheduledTimeoutSender timeoutSender() {
+		return new ScheduledTimeoutSender(Executors.newSingleThreadScheduledExecutor());
+	}
+
+	@Provides
+	@Singleton
+	private FixedTimeoutPacemaker pacemaker(
```
clstrfsck: Return type should be `Pacemaker`, rather than the implementation `FixedTimeoutPacemaker`.

radixdlt/src/main/java/com/radixdlt/consensus/validators/ValidatorSet.java
clstrfsck: No probs.  I can live with this.

radixdlt/src/main/java/com/radixdlt/consensus/validators/Validator.java
```diff
@@ -80,56 +64,32 @@
 	private final boolean jailed;
 */
 
-	@JsonProperty("power")
```
clstrfsck: I think it's a mistake to remove the serialisation stuff here, as we almost certainly will want to persist these at some point.

clstrfsck (changes requested):
The CerberusModule configuration has a bit of untidiness that could be sorted out, having bindings exposed for both Pacemaker and FixedTimeoutPacemaker.

Other than that looks good — a couple of misplaced comments transferred to #79.

radixdlt/src/main/java/com/radixdlt/consensus/liveness/WeightedRotatingLeaders.java
clstrfsck: I think we resolved this offline, but the WRR algorithm produces an output at least every |S| iterations for the sequentially increasing case.  Per our offline discussion, we can revisit this at a later time in any case.

radixdlt/src/main/java/com/radixdlt/consensus/validators/Validator.java
talekhinezh: Even when we do persist these, I would rather have Data object be different from Functional object. Even so, since we're not serializing these yet I think premature to already decide on this class being the serialization mechanism.

clstrfsck (approved):
Looks good to me.

PR #87, opened 2020-04-30
Add utility to output key in base64 format

clstrfsck:


radixdlt/src/main/java/org/radix/OutputKeyBase64.java
```diff
@@ -0,0 +1,69 @@
...
+			if (Modifier.isFinal(isRestricted.getModifiers())) {
+				Field modifiers = Field.class.getDeclaredField("modifiers");
+				modifiers.setAccessible(true);
+				modifiers.setInt(isRestricted, isRestricted.getModifiers() & ~Modifier.FINAL);
+			}
+			isRestricted.setAccessible(true);
+			isRestricted.setBoolean(null, false);
+			isRestricted.setAccessible(false);
+		} catch (ReflectiveOperationException | SecurityException ex) {
+			ex.printStackTrace(System.err);
```
flotothemoon, line 62: Could do with a message here describing what happened

flotothemoon (approved):

radixdlt/src/main/java/org/radix/OutputKeyBase64.java
clstrfsck: OK, context message added.

PR #90, opened 2020-05-08
Fix simulated latency delay

talekhinezh:


radixdlt/src/integration/java/com/radixdlt/consensus/simulation/SimulatedBFTNetwork.java
```diff
@@ -176,8 +180,24 @@ public SystemCounters getCounters(ECKeyPair keyPair) {
...
+		// Send start event once all nodes have reached real epoch event (first epoch event is empty)
+		final CompletableSubject completableSubject = CompletableSubject.create();
+		List<Completable> startedList = this.runners.values().stream()
+			.map(ConsensusRunner::events)
+			.map(o -> Completable.fromSingle(
+				o.map(Event::getEventType)
+					.filter(e -> e.equals(EventType.EPOCH))
+					.skip(1)
+					.firstOrError()
+				)
```
flotothemoon: Might be more idiomatic to express this as `o.map(..)...firstOrError().ignoreElement()` to make it clear that we're intentionally dropping the value here

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/SimulatedBFTNetwork.java
```diff
@@ -176,8 +180,24 @@ public SystemCounters getCounters(ECKeyPair keyPair) {
 		return counters.get(keyPair);
 	}
 
-	public void start() {
+	public Completable start() {
+		// Send start event once all nodes have reached real epoch event (first epoch event is empty)
```
flotothemoon, line 184: Why is the first epoch event empty again? And is it possible (now or ever) that more than just the first event is empty?

talekhinezh: its just to signify that there is no validator set in the beginning of a nodes lifetime.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/SimulatedBFTNetwork.java
talekhinezh: ah! i was looking for `toCompletable`, but `ignoreElement` is what I want :)

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -47,8 +48,7 @@
 	// Should never be null
 	private Vertex root;
 
-	// Should never be null
-	private QuorumCertificate highestQC;
+	private AtomicReference<QuorumCertificate> highestQC = new AtomicReference<>();
```
clstrfsck: What is the thinking here behind using AtomicReference?  The read/update in addQC still has a race condition, so not sure really what the win is over making this volatile?

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -59,7 +59,7 @@ public VertexStore(
 	) {
 		this.engine = Objects.requireNonNull(engine);
 		this.counters = Objects.requireNonNull(counters);
-		this.highestQC = Objects.requireNonNull(rootQC);
+		this.highestQC.set(Objects.requireNonNull(rootQC));
```
clstrfsck: Would be better IMO to assign highestQC with a newly constructed AtomicReference with the reference already in it, eg `this.highestQC = new AtomicReference<>(Objects.requireNonNull(rootQC))`.

That way, if someone removes or otherwise modifies this line, the compilers `final` field initialisation will let them know that there is an issue, also a general dislike of two-phase initialisation when it can be avoided easily.

clstrfsck (approved):

PR #92, opened 2020-05-11
Add stateVersion per vertex

talekhinezh:
To enable easy syncing logic:
Given some committed VertexMetaData, the number of atoms I must download is the difference in stateVersions

### DIFF
--- a/radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
+++ b/radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
@@ -59,7 +59,7 @@ public final class VertexMetadata {
 		}
 
 		this.stateVersion = stateVersion;
-        this.view = view;
+		this.view = view;
 		this.id = id;
 	}
 
@@ -72,7 +72,8 @@ public final class VertexMetadata {
 			parentStateVersion = parent.getStateVersion();
 		}
 
-		final long newStateVersion = parentStateVersion + (vertex.getAtom() != null ? 1 : 0);
+		final int versionIncrement = vertex.getAtom() != null ? 1 : 0;
+		final long newStateVersion = parentStateVersion + versionIncrement;
 		return new VertexMetadata(vertex.getView(), vertex.getId(), newStateVersion);
 	}
 
--- a/radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
+++ b/radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
@@ -42,27 +42,42 @@ public final class VertexMetadata {
 	@DsonOutput(Output.ALL)
 	private final Hash id;
 
+	@JsonProperty("stateVersion")
+	@DsonOutput(Output.ALL)
+	private final long stateVersion;
+
 	VertexMetadata() {
 		// Serializer only
 		this.view = null;
 		this.id = null;
+		this.stateVersion = 0L;
 	}
 
-	public VertexMetadata(View view, Hash id) {
+	public VertexMetadata(View view, Hash id, long stateVersion) {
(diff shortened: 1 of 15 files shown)


flotothemoon (approved):

radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
```diff
@@ -42,27 +42,42 @@
...
 	}
 
-	public VertexMetadata(View view, Hash id) {
+	public VertexMetadata(View view, Hash id, long stateVersion) {
+		if (stateVersion < 0) {
+			throw new IllegalArgumentException("stateVersion must be >= 0");
+		}
+
+		this.stateVersion = stateVersion;
         this.view = view;
```
clstrfsck: Formatto.

radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
```diff
@@ -42,27 +42,42 @@
...
 	VertexMetadata() {
 		// Serializer only
 		this.view = null;
 		this.id = null;
+		this.stateVersion = 0L;
 	}
 
-	public VertexMetadata(View view, Hash id) {
+	public VertexMetadata(View view, Hash id, long stateVersion) {
+		if (stateVersion < 0) {
```
clstrfsck, line 57: If it is an invariant that stateVersion >= 0, then I would suggest adding custom serialization methods (same as view) that actually make this check to ensure that deserialized objects respect this invariant.

radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
```diff
@@ -42,27 +42,42 @@
...
 		if (vertex.isGenesis()) {
-			throw new IllegalStateException("Genesis has no parent.");
+			parentStateVersion = 0;
+		} else {
+			final VertexMetadata parent = vertex.getQC().getProposed();
+			parentStateVersion = parent.getStateVersion();
 		}
 
-		return new VertexMetadata(vertex.getParentView(), vertex.getParentId());
+		final long newStateVersion = parentStateVersion + (vertex.getAtom() != null ? 1 : 0);
```
clstrfsck: Would prefer if we didn't use a ternary operator in this compound expression.

clstrfsck (changes requested):
I think we need to make sure that invariants are respected on deserialization.

A bit concerned that there is no meaningful use of stateVersion.  This makes this PR hard to review, because the intent behind its inclusion is not at all clear.

Also quite concerned that there is no validation of stateVersion, so proposers are free to assert any values they like.  What happens when a proposer proposes something with a value of Long.MAX_LONG (next will be -ve and system broken) or even -1L (next will be zero and numbers reused)?

radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
talekhinezh: It is indeed an invariant but don't think it's a serialization level invariant. Proof of malfeasance would be useful for slashing later.

talekhinezh:
stateVersion is to be used as a simple state accumulator to track how synced we are compared to the latest QC.

proposers don't propose stateVersions, they are constructed by the voter and they wont have any meaning unless a quorum is formed around that stateVersion so it's not tamperable.

radixdlt/src/main/java/com/radixdlt/consensus/VertexMetadata.java
clstrfsck: Ah, OK.  It gets hashed in the VoteData, so should be fine

clstrfsck (approved):

PR #93, opened 2020-05-11
Feature/api return committed only

flotothemoon:
Ensure that only actually committed atoms are returned by the API by changing the `VertexStore` implementation to only store atoms if they were committed (rather than just inserted).

### DIFF
--- a/radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
+++ b/radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
@@ -92,6 +92,11 @@ public final class VertexStore {
 			throw new MissingParentException(vertex.getParentId());
 		}
 
+		// TODO: Don't check for state computer errors for now so that we don't
+		// TODO: have to deal with failing leader proposals
+		// TODO: Reinstate this when ProposalGenerator + Mempool can guarantee correct proposals
+		// TODO: (also see commitVertex->storeAtom)
+
 		vertices.put(vertex.getId(), vertex);
 		updateVertexStoreSize();
 	}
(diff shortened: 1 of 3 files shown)


radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -132,6 +123,20 @@ public Vertex commitVertex(Hash vertexId) {
 		return tipVertex;
 	}
 
+	private void storeAtom(Vertex vertexWithAtom) {
+		try {
+			this.counters.increment(CounterType.LEDGER_PROCESSED);
+			this.engine.store(vertexWithAtom.getAtom());
+			this.counters.increment(CounterType.LEDGER_STORED);
+		} catch (RadixEngineException e) {
+			// TODO: revisit TODO below since this moved from insertVertex to commitVertex
```
flotothemoon: Is the below TODO block still valid? In my view, it still is, though it obviously wouldn't be a vertex insertion failure anymore. @talekhinezh?

talekhinezh: yes, still valid. in future will need this because we still need to "soft-store" atom in engine during prepare phase.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -92,19 +92,6 @@ public void insertVertex(Vertex vertex) throws VertexInsertionException {
 			throw new MissingParentException(vertex.getParentId());
 		}
 
-		if (vertex.getAtom() != null) {
```
talekhinezh, line 95: Need to add a comment here saying the current hack of automatically accepting anything in the prepare phase.

flotothemoon: Good point, done

PR #97, opened 2020-05-25
Load validator constraint scrypt into middleware

flotothemoon:
Title, basically. Loads validator constraint scrypt (which was recently added to the RE) into the middleware to make it available in core.


flotothemoon:
Closing because it's duplicated.

PR #102, opened 2020-05-28
Feature/sync ledger request

talekhinezh:
* Added SyncedStateComputer interface
* Implemented Vertex retrieval
* Implemented Committed state retrieval (though not yet used)
* Added BFT.stop json rpc method for sync testing
* Added One Proposal Per View Dropper deterministic + simulation tests


radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -17,168 +17,288 @@
...
 		}
 
-		for (Vertex committed : path) {
-			if (committed.getAtom() != null) {
-				CommittedAtom committedAtom = committed.getAtom().committed(commitMetadata);
-				storeAtom(committedAtom);
-			}
 
-			lastCommittedVertex.onNext(committed);
+		/*
```
flotothemoon: Why is this commented out? Can we remove it?

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -101,66 +114,95 @@ public Builder randomLatency(int minLatency, int maxLatency) {
...
+				BFTCheck check = this.checks.get(name);
+				return
+					Pair.of(
+						name,
+						check.check(bftNetwork).map(e -> Pair.of(name, e)).publish().autoConnect(2)
+					);
+			})
+			.collect(Collectors.toList());
+
+		Single<String> doneSignal = Observable.merge(assertions.stream().map(Pair::getSecond).collect(Collectors.toList()))
```
flotothemoon: Just looking at this code, it isn't clear to me what it does or how it works. What is `done` (success? failure? either?)  and how does the name of the first check relate to it?

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -101,66 +114,95 @@ public Builder randomLatency(int minLatency, int maxLatency) {
...
+			.firstOrError()
+			.map(Pair::getFirst);
+
+		List<Single<Pair<String, Boolean>>> results = assertions.stream()
+			.map(assertion -> assertion.getSecond()
+				.takeUntil(doneSignal.filter(done -> !assertion.getFirst().equals(done)).toObservable().concatWith(Observable.never()))
```
flotothemoon: What is the `.concatWith(Observable.never())` for here?

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -101,66 +114,95 @@ public Builder randomLatency(int minLatency, int maxLatency) {
...
-	public void run(long duration, TimeUnit timeUnit) {
+	/**
+	 * Runs the test for a given time. Returns either once the duration has passed or if a check has failed.
+	 * Returns a map from the check name to the result.
+	 *
+	 * @param duration duration to run test for
+	 * @param timeUnit time unit of duration
+	 * @return map of check results
+	 */
+	public Map<String, Boolean> run(long duration, TimeUnit timeUnit) {
```
flotothemoon, line 170: I like this change to a more expressive return value. It could be made even more expressive if we were to include `BFTCheckError` or a similar structure describing the result rather than just a boolean value

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/OneProposalPerViewDropper.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public class OneProposalPerViewDropper implements Predicate<MessageInTransit> {
+	private final Cache<View, ECPublicKey> proposalToDrop = CacheBuilder.newBuilder()
+		.maximumSize(100)
+		.build();
+
+	private final ImmutableList<ECPublicKey> nodes;
+	private final Random random;
+
+	public OneProposalPerViewDropper(ImmutableList<ECPublicKey> nodes, Random random) {
```
flotothemoon, line 39: Is there a specific reason the node selection has to be random? 
If not, might be easier to turn the random selection into a general function that can be passed in.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/OneProposalPerViewDropper.java
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	@Override
+	public boolean test(MessageInTransit msg) {
+		if (msg.getContent() instanceof Proposal) {
+			final Proposal proposal = (Proposal) msg.getContent();
+			final View view = proposal.getVertex().getView();
+			final ECPublicKey nodeToDrop;
+			try {
+				nodeToDrop = proposalToDrop.get(view, () -> nodes.get(random.nextInt(nodes.size())));
```
flotothemoon: Not sure this works as intended given that random will return a different value every time its called, yielding a different `nodeToDrop` if the cache overruns its 100 capacity. Setting a seed here might fix this, but still seems fragile (i.e. if we need to set a seed here, why pass in a random instance at all?).

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/OneProposalPerViewDropper.java
```diff
@@ -0,0 +1,66 @@
...
+			} catch (ExecutionException e) {
+				throw new IllegalStateException(e);
```
flotothemoon: An error message here would be good

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/LivenessCheck.java
```diff
@@ -53,12 +49,16 @@ public Completable check(SimulatedBFTNetwork network) {
...
 				.max(View::compareTo)
-				.get()) // there must be some max highest QC unless allNodes is empty
-			.doOnNext(view -> assertThat(view)
-				.satisfies(new Condition<>(v -> v.compareTo(highestQCView.get()) > 0,
-					"The highest highestQC %s increased since last highestQC %s after %d %s", view, highestQCView.get(), duration, timeUnit)))
-			.doOnNext(highestQCView::set)
-			.doOnNext(newHighestQCView -> System.out.println("Progressed to new highest QC view " + highestQCView))
-			.flatMapCompletable(v -> Completable.complete());
+				.get()
+			) // there must be some max highest QC unless allNodes is empty
```
flotothemoon: the comment here refers to the `get()` on the `Optional`, so it should be on the line above

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/LivenessCheck.java
```diff
@@ -53,12 +49,16 @@ public Completable check(SimulatedBFTNetwork network) {
...
+			.concatMap(view -> {
+				if (view.compareTo(highestQCView.get()) <= 0) {
+					return Observable.just(new BFTCheckError("Highest QC hasn't increased"));
```
flotothemoon: Adding more details here would make debugging easier, e.g. what was the last highest QC, what's the current, time elapsed, etc.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/NoSyncExceptionCheck.java
```diff
@@ -20,24 +20,24 @@
...
+	public Observable<BFTCheckError> check(SimulatedBFTNetwork network) {
 		return Observable.interval(1, TimeUnit.SECONDS)
 			.flatMapIterable(i -> network.getNodes())
-			.doOnNext(cn -> AssertionsForClassTypes.assertThat(network.getCounters(cn).get(CounterType.CONSENSUS_SYNC_EXCEPTION))
-				.satisfies(new Condition<>(c -> c == 0,
-					"Sync Exception counter is zero in correct node %s", cn.getPublicKey().euid())))
-			.flatMapCompletable(p -> Completable.complete());
+			.concatMap(cn -> {
+				if (network.getCounters(cn).get(CounterType.CONSENSUS_SYNC_EXCEPTION) > 0) {
+					return Observable.just(new BFTCheckError("Sync Exception Count > 0"));
```
flotothemoon: Please include the actual value in the error

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/NoTimeoutCheck.java
```diff
@@ -34,16 +31,16 @@
...
-			.doOnNext(counters -> {
-				assertThat(counters.get(CounterType.CONSENSUS_TIMEOUT))
-					.satisfies(new Condition<>(c -> c == 0, "Timeout counter is zero."));
-				assertThat(counters.get(CounterType.CONSENSUS_REJECTED))
-					.satisfies(new Condition<>(c -> c == 0, "Rejected Proposal counter is zero."));
-			})
-			.flatMapCompletable(c -> Completable.complete());
+			.concatMap(counters -> {
+				if (counters.get(CounterType.CONSENSUS_TIMEOUT) > 0) {
+					return Observable.just(new BFTCheckError("Timeout counter > 0"));
```
flotothemoon: Please include the actual value in the error

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/SafetyCheck.java
```diff
@@ -33,15 +30,21 @@
...
 				.collect(Collectors.toList()),
-			Arrays::stream)
+				Arrays::stream
+		)
 			.map(committedVertices -> committedVertices.distinct().collect(Collectors.toList()))
-			.doOnNext(committedVertices -> assertThat(committedVertices).hasSize(1))
-			.flatMapCompletable(v -> Completable.complete());
+			.concatMap(committedVertices -> {
+				if (committedVertices.size() != 1) {
+					return Observable.just(new BFTCheckError("Committed vertices don't match"));
```
flotothemoon: Please include the vertex IDs in the error msg

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/OneProposalDropperTest.java
```diff
@@ -0,0 +1,74 @@
...
+	 */
+	@Test
+	public void given_get_vertices_enabled__then_test_should_succeed_against_drop_proposal_adversary() {
+		BFTSimulatedTest test = bftTestBuilder
+			.setGetVerticesRPCEnabled(true)
+			.build();
+
+		Map<String, Boolean> results = test.run(1, TimeUnit.MINUTES);
+		System.out.println(results);
+		assertThat(results).doesNotContainValue(false);
```
flotothemoon, line 70: Nice, that's pretty slick.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/OneSlowNodeTest.java
```diff
@@ -35,16 +38,20 @@
...
 		.numNodesAndLatencies(4, minLatency, minLatency, minLatency, maxLatency)
 		.pacemakerTimeout(synchronousTimeout)
-		.checkNoTimeouts();
+		.checkSafety("safety")
+		.checkAllProposalsHaveDirectParents("directParents")
+		.checkNoTimeouts("noTimeouts");
 
 	/**
 	 * Tests a static configuration of 3 fast, equal nodes and 1 slow node.
+	 * Test should pass even with GetVertices RPC disabled
```
flotothemoon, line 47: Why should the test pass without syncing, given that there's a lagging node?

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/RandomLatencyTest.java
```diff
@@ -28,54 +32,49 @@
...
+	// a round can consist of 6 * max_transmission_time (MTT)
+	private final int trips = 6;
+	private final int synchronousTimeout = maxLatency * trips;
+
+	private Builder bftTestBuilder = BFTSimulatedTest.builder()
+		.randomLatency(minLatency, maxLatency)
+		.setGetVerticesRPCEnabled(false)
+		.pacemakerTimeout(synchronousTimeout) // Since no syncing needed 6*MTT required
+		.checkLiveness("liveness", synchronousTimeout, TimeUnit.MILLISECONDS)
+		.checkSafety("safety")
```
flotothemoon, line 49: Seeing these strings repeated very often makes me think that a default overload in BFTSimulatedTest builder for these common checks would be good

radixdlt/src/integration/java/com/radixdlt/store/berkeley/BerkeleyRadixLedgerEntryStoreTests.java
```diff
@@ -39,107 +38,110 @@
...
 
             //added atom is present in store
-            softly.assertThat(tempoAtomStore.contains(ledgerEntries.get(0).getAID())).isTrue();
+            softly.assertThat(ledgerStore.contains(ledgerEntries.get(0).getAID())).isTrue();
 
             // commit atom
-            tempoAtomStore.commit(ledgerEntries.get(0).getAID());
+            ledgerStore.commit(ledgerEntries.get(0).getAID());
+
+            softly.assertThat(ledgerStore.getNextCommitted(ledgerEntries.get(0).getStateVersion() - 1, 1))
```
flotothemoon, line 81: Can you explain what this checks for?

flotothemoon: (and maybe add a comment in-code too)

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -17,168 +17,288 @@
...
-	public Vertex commitVertex(VertexMetadata commitMetadata) {
-		final Hash vertexId = commitMetadata.getId();
-		final Vertex tipVertex = vertices.get(vertexId);
-		if (tipVertex == null) {
-			throw new IllegalStateException("Committing a vertex which was never inserted: " + vertexId);
+	public void processLocalSync(Hash vertexId) {
+		syncing.remove(vertexId);
+	}
+
+	public boolean syncToQC(QuorumCertificate qc, QuorumCertificate committedQC, @Nullable ECPublicKey author) {
```
flotothemoon, line 176: Would we ever want to sync on any `committedQC` that's not the highest QC of this vertex store?

flotothemoon: Regardless, `committedQC` appears to be unused anyway, so should either use or remove.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -17,168 +17,288 @@
...
 
-		updateVertexStoreSize();
-		return tipVertex;
+			QuorumCertificate highestCommitted = highestCommittedQC.get();
+			qc.getCommitted().ifPresent(vertexMetadata -> {
+				if (!highestCommitted.getCommitted().isPresent()
+					|| highestCommitted.getCommitted().get().getStateVersion() < vertexMetadata.getStateVersion()) {
+					this.highestCommittedQC.set(qc);
+				}
+			});
```
flotothemoon: For an atomic version, this could be turned into something like:
```suggestion
			QuorumCertificate highestCommitted = highestCommittedQC.get();
			highestCommittedQC.getAndUpdate(currentHighest -> qc.getCommitted().map(vertexMetadata -> {
			if (!highestCommitted.getCommitted().isPresent()
				|| highestCommitted.getCommitted().get().getStateVersion() < vertexMetadata.getStateVersion()) {
				return qc;
			} else {
				return currentHighest;
			}
		}).orElse(currentHighest));
```

flotothemoon: Urgh, the formatting of the suggested change here is a bit awful. My suggestion is essentially just moving the get + update wholly inside AtomicReference's own getAndUpdate.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -187,33 +307,58 @@ private void storeAtom(CommittedAtom atom) {
...
 	 *
 	 * @return the highest quorum certificate
 	 */
 	public QuorumCertificate getHighestQC() {
-		return this.highestQC;
+		return this.highestQC.get();
 	}
 
 	/**
 	 * Retrieves the vertex with the given vertexId if it exists in the store.
```
flotothemoon: Documentation needs updating to reflect interface/behaviour change

flotothemoon (commented):
Some initial comments. Part 2 coming soon, stay tuned.

radixdlt/src/main/java/com/radixdlt/consensus/ConsensusRunner.java
```diff
@@ -65,15 +71,20 @@ public String toString() {
...
 	@Inject
 	public ConsensusRunner(
 		EpochRx epochRx,
 		EventCoordinatorNetworkRx network,
 		PacemakerRx pacemakerRx,
-		EpochManager epochManager
+		LocalSyncRx localSyncRx,
+		SyncVerticesRPCRx rpcRx,
+		EpochManager epochManager,
+		VertexStore vertexStore //TODO: remove this since it should only be provided by Epoch manager
```
flotothemoon, line 85: Why should the vertex store be provided by the epoch manager?

radixdlt/src/main/java/com/radixdlt/consensus/NewView.java
```diff
@@ -58,15 +62,23 @@
 		this.author = null;
 		this.view = View.of(Long.MAX_VALUE);
 		this.signature = null;
+		this.committedQC = null;
 	}
 
-	public NewView(ECPublicKey author, View view, QuorumCertificate qc, ECDSASignature signature) {
+	public NewView(ECPublicKey author, View view, QuorumCertificate qc, QuorumCertificate commitedQC, ECDSASignature signature) {
```
flotothemoon: `commitedQC` => `committedQC`

radixdlt/src/main/java/com/radixdlt/consensus/SyncVerticesRPCRx.java
```diff
@@ -0,0 +1,39 @@
...
+import com.radixdlt.consensus.VertexStore.GetVerticesRequest;
+import io.reactivex.rxjava3.core.Observable;
+
+/**
+ * Provider of GetVertices RPC request/response events
+ */
+public interface SyncVerticesRPCRx {
+
+	/**
+	 * Retrieve a never-ending stream of requests
```
flotothemoon, line 29: `requests` and `responses` are hot observables, right?

radixdlt/src/main/java/com/radixdlt/consensus/sync/StateSyncNetwork.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.List;
+
+/**
+ * Network interface for state syncing
+ */
+public interface StateSyncNetwork {
+
+	/**
+	 * Retrieve stream of sync responses
+	 * @return an unending Observable of sync responses
```
flotothemoon, line 32: `syncResponses` and `syncRequests` are hot as well, right?

radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesRequestMessage.java
```diff
@@ -28,27 +28,37 @@
...
+@SerializerId2("message.consensus.vertices_request")
+public final class GetVerticesRequestMessage extends Message {
 	@JsonProperty("vertexId")
 	@DsonOutput(Output.ALL)
 	private final Hash vertexId;
 
-	GetVertexRequestMessage() {
+	@JsonProperty("count")
+	@DsonOutput(Output.ALL)
+	private final int count;
```
flotothemoon, line 39: Is this a required count or more of a limit?

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralSyncCommittedNetwork.java
```diff
@@ -0,0 +1,78 @@
...
+public final class MessageCentralSyncCommittedNetwork implements StateSyncNetwork {
+	private final int magic;
+	private final MessageCentral messageCentral;
+
+	@Inject
+	public MessageCentralSyncCommittedNetwork(
+		Universe universe,
+		MessageCentral messageCentral
+	) {
+		this.magic = universe.getMagic();
```
flotothemoon, line 45: Can we pass in magic directly? Or will we need universe for something?

radixdlt/src/main/java/com/radixdlt/consensus/GetVerticesResponse.java
```diff
@@ -18,31 +18,37 @@
...
+import java.util.List;
+import java.util.Objects;
 
 /**
- * An RPC request to retrieve a given vertex
+ * An RPC response
  */
-public final class GetVertexRequest {
+public final class GetVerticesResponse {
+	private final Object opaque;
```
flotothemoon, line 28: Why is `opaque` opaque rather than just a plain `Hash`?

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
```diff
@@ -0,0 +1,173 @@
...
+	private final StateSyncNetwork stateSyncNetwork;
+
+	@Inject
+	public SyncedRadixEngine(
+		RadixEngine<LedgerAtom> radixEngine,
+		CommittedAtomsStore committedAtomsStore,
+		AddressBook addressBook,
+		StateSyncNetwork stateSyncNetwork
+	) {
+		this.radixEngine = radixEngine;
```
flotothemoon: Should check for nullness here

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
```diff
@@ -0,0 +1,173 @@
...
+	 * given a target list and then requests for atom inventory.
+	 *
+	 * TODO: cancel previous unfinished requests
+	 *
+	 * @param targetStateVersion the target state version
+	 * @param target a list of potential targets
+	 * @return completable which completes when the store has completed syncing
+	 */
+	@Override
+	public Completable syncTo(long targetStateVersion, List<ECPublicKey> target) {
```
flotothemoon, line 116: Should we check if target is empty here to avoid confusing errors? Or maybe just add the given number of possible peers to the error message below

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		Peer peer = target.stream()
+			.map(pk -> addressBook.peer(pk.euid()))
+			.filter(Optional::isPresent)
+			.map(Optional::get)
+			.filter(Peer::hasSystem)
+			.findFirst()
+			.orElseThrow(() -> new RuntimeException("Unable to find peer"));
+		stateSyncNetwork.sendSyncRequest(peer, currentStateVersion);
```
flotothemoon, line 135: If a node is lagging behind a bit, it may issue many requests to whoever the first viable peer in `target` is before actually catching up (because requests don't coalesce). Is this something we want to change later or do you think it's not an issue? 
If we want to come back to it later, we may at least want to randomly pick our peer from `target`.

radixdlt/src/main/java/com/radixdlt/middleware2/store/CommittedAtomsStore.java
```diff
@@ -96,7 +105,13 @@ public void storeAtom(CommittedAtom committedAtom) {
         VertexMetadata vertexMetadata = committedAtom.getVertexMetadata();
         LedgerEntry ledgerEntry = new LedgerEntry(binaryAtom, vertexMetadata.getStateVersion(), committedAtom.getAID());
         EngineAtomIndices engineAtomIndices = atomIndexer.getIndices(committedAtom);
+
+        // TODO: Replace Store + Commit with a single commit
+		// TODO: How it's done depends on how mempool and prepare phases are implemented
```
flotothemoon, line 110: Odd extra indent here

radixdlt/src/main/java/com/radixdlt/middleware2/store/CommittedAtomsStore.java
```diff
@@ -109,6 +124,32 @@ public void storeAtom(CommittedAtom committedAtom) {
         lastStoredAtom.onNext(storedEvent);
     }
 
+    /**
+     * Retrieve the committed atoms in the store after a given state version
+     * @param stateVersion the state version on which we will return atoms which are after
+     * @param limit limit to number of atoms to return
```
flotothemoon: starting at state version + 1 (inclusively)?

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
talekhinezh: Yes, totally agree but would rather have that be a task for the future. It doesn't gain us anything immediately.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
talekhinezh: The idea is to have the test run either until timeout or until one of the tests fail an assertion at which point we collect all current results of each test.

`done` should probably be changed to `firstError`. The name is used so that we don't stop the error check from emitting it's error to the second stream below.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
talekhinezh: this is so that if the doneSignal is equal to the check that the `.takeUntil` doesn't get signaled

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/OneProposalPerViewDropper.java
talekhinezh: No, but would rather not add indirection until there are more tests which will use it.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/OneProposalPerViewDropper.java
talekhinezh: err yeah this is a bit of a hack, the idea is that we can assume that a proposal for view v wont be sent anyone by the time we get to view v+100.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/OneSlowNodeTest.java
talekhinezh: Just -relatively- lagging. All nodes still receive all messages from all nodes within synchrony bounds.

radixdlt/src/main/java/com/radixdlt/consensus/ConsensusRunner.java
talekhinezh: A BFT run with fixed validatorSet from start to end is an "epoch". Thus VertexStore which is specifically used by BFT should have a lifecycle maintained by the epoch.

radixdlt/src/main/java/com/radixdlt/consensus/GetVerticesResponse.java
talekhinezh: Because it contains information only relevant to the requestor for management purposes and nothing to do with the request/response itself

radixdlt/src/main/java/com/radixdlt/consensus/SyncVerticesRPCRx.java
talekhinezh: no, cold observables but doesn't really matter

radixdlt/src/main/java/com/radixdlt/consensus/sync/StateSyncNetwork.java
talekhinezh: ice cold

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
talekhinezh: yeah definitely we want to make better for future

radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesRequestMessage.java
talekhinezh: required

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralSyncCommittedNetwork.java
talekhinezh: no dont need it for anything. Should actually just aim to remove it as I believe `magic` isn't protected by any signature.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
talekhinezh: This has been improved slightly in the next PR

radixdlt/src/integration/java/com/radixdlt/store/berkeley/BerkeleyRadixLedgerEntryStoreTests.java
talekhinezh: that a committed atom can be retrieved by version

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: To be used in next PR. Had to make a bit of a U-turn during development (also reason why this PR is so big).

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: Used in the next PR. And yes, we need highest committed in order to know how to sync the committed state.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: this no longer needs to be atomically updated since only one thread is ever accessing VertexStore at a time.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/RandomLatencyTest.java
talekhinezh: I think that just hides what we want to test.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
flotothemoon: Hmm okay, does it need to be here already then if it's not used yet? It's not like it's a feature flag or similar, it's just commented code we think we'll need.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/OneSlowNodeTest.java
flotothemoon: That would be great information to have in the comment above :)

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/OneProposalPerViewDropper.java
flotothemoon: I guess that's a fine assumption for simulated tests as long as we're aware & it's documented.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/RandomLatencyTest.java
flotothemoon: Maybe I wasn't being clear. My proposal is not to move any of the checks to the builder, but just to have there be a default for the check *name*, like this:
```suggestion
		.checkSafety()
```

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
flotothemoon: Cool, ack

radixdlt/src/main/java/com/radixdlt/consensus/LocalSyncRx.java
```diff
@@ -17,22 +17,18 @@
...
-	 * TODO: refactor to maintain a unidirectional data flow
-	 *
-	 * @param id the id of the vertex to retrieve
-	 * @param node the node to retrieve the vertex info from
-	 * @return single of a vertex which will complete once retrieved
+	 * Retrieve rx flow of vertex hashes
+	 * @return flow of vertex hashes
 	 */
-	Single<Vertex> getVertex(Hash id, ECPublicKey node);
+	Observable<Hash> localSyncs();
```
flotothemoon, line 33: What makes these 'local'?

radixdlt/src/main/java/com/radixdlt/consensus/SyncVerticesRPCRx.java
flotothemoon: not sure why this doesn't matter, isn't that relevant to know as a consumer of this interface?

radixdlt/src/main/java/com/radixdlt/consensus/sync/StateSyncNetwork.java
flotothemoon: latte

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -17,92 +17,235 @@
...
+		private final LinkedList<Vertex> fetched = new LinkedList<>();
+
+		SyncState(QuorumCertificate qc, ECPublicKey author) {
+			this.qc = qc;
+			this.author = author;
+		}
+	}
+
+	public void processGetVerticesRequest(GetVerticesRequest request) {
+		log.info("SYNC_VERTICES: Received GetVerticesRequest {}", request);
```
flotothemoon, line 112: This is definitely a place where we'll want to exercise care when we move into Byzantine land (consider nodes sending ridiculous requests, forcing us to gather and send a bunch of vertices). Maybe it should be marked as such for later consideration, what do you think?

flotothemoon (approved):
Sparing some outstanding comments, this is good to go by me.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -101,66 +114,96 @@ public Builder randomLatency(int minLatency, int maxLatency) {
...
-
-		public Builder checkSyncsHaveNotOccurred(long time, TimeUnit timeUnit) {
-			this.checks.add(new SyncOccurrenceCheck(c -> assertThat(c).isEqualTo(0), time, timeUnit));
-			return this;
-		}
-
-		public Builder checkLiveness() {
-			this.checks.add(new LivenessCheck(8 * TestEventCoordinatorNetwork.DEFAULT_LATENCY, TimeUnit.MILLISECONDS));
+		public Builder checkLiveness(String checkName) {
+			this.checks.put(checkName, new LivenessCheck(8 * TestEventCoordinatorNetwork.DEFAULT_LATENCY, TimeUnit.MILLISECONDS));
```
clstrfsck: Probably should be some kind of exception if something gets overwritten by a put in all of these methods.  Pretty bad if it's believed that some kind of check is running, but it's been silently overwritten because of poor name choice.

Maybe even consider an ImmutableMap builder, which I think does these checks and throws.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -18,57 +18,70 @@
...
 		this.latencyProvider = latencyProvider;
 		this.checks = checks;
 		this.pacemakerTimeout = pacemakerTimeout;
+		this.getVerticesRPCEnabled = getVerticesRPCEnabled;
 	}
 
 	public static class Builder {
 		private final DroppingLatencyProvider latencyProvider = new DroppingLatencyProvider();
-		private final List<BFTCheck> checks = new ArrayList<>();
+		private final Map<String, BFTCheck> checks = new HashMap<>();
```
clstrfsck: Use of a HashMap here makes the ordering of checks unpredictable.  Better would be ImmutableMap builder.

See also comments about checks being overwritten.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -101,66 +114,96 @@ public Builder randomLatency(int minLatency, int maxLatency) {
...
-		public Builder checkLiveness(long duration, TimeUnit timeUnit) {
-			this.checks.add(new LivenessCheck(duration, timeUnit));
+		public Builder checkLiveness(String checkName, long duration, TimeUnit timeUnit) {
+			this.checks.put(checkName, new LivenessCheck(duration, timeUnit));
 			return this;
 		}
 
-		public Builder checkSafety() {
-			this.checks.add(new SafetyCheck());
+		public Builder checkSafety(String checkName) {
```
clstrfsck, line 126: Not sure why this method, and others that can only be applied once to a test, don't have a standard, fixed name.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/RandomLatencyTest.java
clstrfsck: I agree with this.  There's no meaningful use case for adding the safety check twice right now (as well as others), so would be more consistent and less error prone if it was added with a fixed name/key.

radixdlt/src/main/java/com/radixdlt/consensus/LocalSyncRx.java
talekhinezh: It's a message originating from a local module

radixdlt/src/main/java/com/radixdlt/consensus/ConsensusRunner.java
```diff
@@ -92,24 +103,29 @@ public ConsensusRunner(
...
 
 		this.events = Observable.merge(epochs, ecMessages).publish();
 	}
 
 	private Event processEvent(Object msg, BFTEventProcessor processor) {
 		final EventType eventType;
-		if (msg instanceof GetVertexRequest) {
-			processor.processGetVertexRequest((GetVertexRequest) msg);
-			return new Event(EventType.GET_VERTEX_REQUEST, msg);
+		if (msg instanceof GetVerticesRequest) {
```
clstrfsck, line 123: This is getting pretty bad now, and needs a rethink.

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
```diff
@@ -0,0 +1,180 @@
...
+import org.radix.atoms.particles.conflict.ParticleConflictException;
+import org.radix.events.Events;
+import org.radix.validation.ConstraintMachineValidationException;
+
+/**
+ * A service which synchronizes the radix engine committed state between peers.
+ *
+ * TODO: Most of the logic here should go into RadixEngine itself
+ */
+@Singleton
```
clstrfsck, line 52: Would prefer if this was flagged singleton in the module that used it, rather than the class itself.

Ideally the property of being a singleton is a consequence of how the class is used, rather than its implementation.

radixdlt/src/main/java/com/radixdlt/consensus/LocalSyncSender.java
```diff
@@ -0,0 +1,50 @@
...
+import com.radixdlt.consensus.VertexStore.SyncSender;
+import com.radixdlt.crypto.Hash;
+import io.reactivex.rxjava3.core.Observable;
+import io.reactivex.rxjava3.subjects.BehaviorSubject;
+import io.reactivex.rxjava3.subjects.Subject;
+
+/**
+ * Sends a local sync event and exposes it as a stream of sync events
+ */
+@Singleton
```
clstrfsck, line 31: Would prefer if this was flagged singleton in the module that used it, rather than the class itself.

Ideally the property of being a singleton is a consequence of how the class is used, rather than its implementation.

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralBFTNetwork.java
```diff
@@ -52,7 +48,8 @@
  * BFT Network sending and receiving layer used on top of the MessageCentral
  * layer.
  */
-public final class MessageCentralBFTNetwork implements BFTEventSender, EventCoordinatorNetworkRx, VertexSupplier {
+@Singleton
```
clstrfsck, line 51: Would prefer if this was flagged singleton in the module that used it, rather than the class itself.

Ideally the property of being a singleton is a consequence of how the class is used, rather than its implementation.

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralSyncVerticesRPCNetwork.java
```diff
@@ -0,0 +1,158 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
+
+/**
+ * Network interface for syncing vertices using the MessageCentral
+ */
+@Singleton
```
clstrfsck: Would prefer if this was flagged singleton in the module that used it, rather than the class itself.

Ideally the property of being a singleton is a consequence of how the class is used, rather than its implementation.

radixdlt/src/main/java/com/radixdlt/consensus/GetVerticesResponse.java
```diff
@@ -18,31 +18,37 @@
...
-		this.responder = responder;
+	public GetVerticesResponse(Hash vertexId, List<Vertex> vertices, Object opaque) {
+		this.vertexId = Objects.requireNonNull(vertexId);
+		this.vertices = Objects.requireNonNull(vertices);
+		this.opaque = opaque;
 	}
 
-	public Consumer<Vertex> getResponder() {
-		return responder;
+	public <T> T getOpaque(Class<T> c) {
```
clstrfsck: Why is this cast here?  It looks to me like this should be the responsibility of client code to do the cast, given that the opaque object was supplied by the same code?

radixdlt/src/main/java/com/radixdlt/middleware2/store/CommittedAtomsStore.java
```diff
@@ -96,7 +105,13 @@ public void storeAtom(CommittedAtom committedAtom) {
         VertexMetadata vertexMetadata = committedAtom.getVertexMetadata();
         LedgerEntry ledgerEntry = new LedgerEntry(binaryAtom, vertexMetadata.getStateVersion(), committedAtom.getAID());
         EngineAtomIndices engineAtomIndices = atomIndexer.getIndices(committedAtom);
+
+		// TODO: Replace Store + Commit with a single commit
```
clstrfsck, line 109: This file looks like it's using a mixture of tabs and spaces.

radixdlt/src/main/java/org/radix/Radix.java
```diff
@@ -142,6 +143,9 @@ public static void start(RuntimeProperties properties) {
 		// Start mempool receiver
 		globalInjector.getInjector().getInstance(MempoolReceiver.class).start();
 
+		SyncedRadixEngine stateSynchronizer = globalInjector.getInjector().getInstance(SyncedRadixEngine.class);
```
clstrfsck: I don't really like this, or the mempool receiver handling, but I can't think of anything immediately better.

clstrfsck (commented):
Some suggestions, at least some of them should be addressed IMO.

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralSyncCommittedNetwork.java
clstrfsck: All of the elements used to compute the magic are signed by the universe creation key — they are fields in the universe, and the universe is signed.

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralSyncVerticesRPCNetwork.java
```diff
@@ -0,0 +1,158 @@
...
+public class MessageCentralSyncVerticesRPCNetwork implements SyncVerticesRPCSender, SyncVerticesRPCRx {
+
+	private final ECPublicKey selfPublicKey;
+	private final int magic;
+	private final AddressBook addressBook;
+	private final MessageCentral messageCentral;
+	private Cache<Hash, Object> opaqueCache = CacheBuilder.newBuilder()
+		.expireAfterWrite(30, TimeUnit.SECONDS)
```
clstrfsck, line 53: Timeout really should be configurable.

Also opaqueCache should be final.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/synchronous/RandomLatencyTest.java
talekhinezh: Have to disagree here. It's less error prone if you can see upfront what test names you are using, especially if you are going to use those same names to check for specific results. It's not about use cases at this level but more about readability (like builder patterns).

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
talekhinezh: yup sounds good

radixdlt/src/main/java/com/radixdlt/consensus/ConsensusRunner.java
talekhinezh: Yes, but for a future PR i think

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralBFTNetwork.java
talekhinezh: good point

radixdlt/src/main/java/org/radix/Radix.java
talekhinezh: What's the issue?

radixdlt/src/main/java/com/radixdlt/consensus/GetVerticesResponse.java
talekhinezh: yup good point

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: yup

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralSyncVerticesRPCNetwork.java
talekhinezh: Gonna hold off on making this configurable until we settle on right approach for handling RPCs. Don't want to prematurely overdesign here.

PR #103, opened 2020-06-02
Committed Sync

talekhinezh:


radixdlt/src/main/java/com/radixdlt/middleware2/network/GetVerticesResponseMessage.java
```diff
@@ -31,23 +33,33 @@
...
 		this.vertices = Objects.requireNonNull(vertices);
 	}
 
+	public Hash getVertexId() {
+		return vertexId;
+	}
+
 	public List<Vertex> getVertices() {
-		return vertices;
+		return vertices == null ? Collections.emptyList() : vertices;
```
clstrfsck: Would prefer this to be returning an immutable list, so that this object remained immutable.

clstrfsck (dismissed):

clstrfsck (changes requested):
Looks OK, but tests are failing for some reason.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -167,7 +168,7 @@ public static Builder builder() {
 	 * @param timeUnit time unit of duration
 	 * @return map of check results
 	 */
-	public Map<String, Boolean> run(long duration, TimeUnit timeUnit) {
+	public Map<String, Optional<BFTCheckError>> run(long duration, TimeUnit timeUnit) {
```
flotothemoon, line 171: Yay!

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/SafetyCheck.java
```diff
@@ -17,34 +17,105 @@
...
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 /**
- * Checks that nodes are committed on the same vertex.
+ * Checks that nodes do not commit on conflicting vertices
  */
 public class SafetyCheck implements BFTCheck {
 
+	private static Observable<BFTCheckError> differentVerticesError(Vertex vertex, Vertex currentVertex) {
```
flotothemoon: "conflicting" might be more clear

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/SafetyCheck.java
```diff
@@ -17,34 +17,105 @@
...
+					highest.set(vertex.getView());
 				}
+
+				// Clean up old vertices so that we avoid consuming too much memory
+				lastCommittedByNode.put(node, vertex);
+				final View lowest = lastCommittedByNode.values().stream()
+					.map(Vertex::getView)
+					.reduce((v0, v1) -> v0.compareTo(v1) < 0 ? v0 : v1)
+					.orElse(View.genesis());
+				final Set<View> viewsToRemove = committedVertices.keySet().stream()
```
Sajjon, line 114: Skip using sets and use array instead to get deterministic behaviour? Or is it important to have non-determinism here? :) For uniqueness you can use `.distinct()` I think?

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/SafetyCheck.java
```diff
@@ -17,34 +17,105 @@
...
+				node -> network.getVertexStore(node)
+					.lastCommittedVertex().map(v -> Pair.of(node, v)))
+				.collect(Collectors.toList())
 		)
-			.map(committedVertices -> committedVertices.distinct().collect(Collectors.toList()))
-			.concatMap(committedVertices -> {
-				if (committedVertices.size() != 1) {
-					return Observable.just(new BFTCheckError("Committed vertices don't match: " + committedVertices));
+			.flatMap(nodeAndVertex -> {
+				final ECKeyPair node = nodeAndVertex.getFirst();
```
Sajjon, line 86: Pretty long anonymous function inside this `flatMap`, making it hard to follow the reactive code, maybe break out to helper method? :)

radixdlt/src/main/java/com/radixdlt/consensus/SyncedStateComputer.java
```diff
@@ -47,9 +46,11 @@
 	 *
 	 * @param targetStateVersion the target state version
 	 * @param target list of targets as hint of which peer has the state
+	 * @param opaque some opaque client object which will be passed in a sync
+	 * message if this returns false
 	 * @return a completable which completes when sync is complete
```
flotothemoon: update documentation, no longer a completable

Sajjon (commented):
When Rx anonymous functions become too long it becomes hard to follow the reactive flow, when possible, I think it would be advantageous to break out helper method, what do you think?

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -115,73 +182,164 @@ public void processGetVerticesRequest(GetVerticesRequest request) {
...
 
-	private void doSync(QuorumCertificate qc, ECPublicKey author) {
-		final Hash vertexId = qc.getProposed().getId();
+	public void processGetVerticesResponse(GetVerticesResponse response) {
+		log.info("SYNC_VERTICES: Received GetVerticesResponse {}", response);
+
+		final Hash syncTo = (Hash) response.getOpaque();
+		SyncState syncState = syncing.get(syncTo);
+		if (syncState == null) {
+			return;
```
flotothemoon: Does this signal an invalid or malicious response? Can we do something here besides ignoring it (e.g. log it)?

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -115,73 +182,164 @@ public void processGetVerticesRequest(GetVerticesRequest request) {
...
 		if (addQC(qc)) {
 			return true;
 		}
 
-		if (author != null) {
-			this.doSync(qc, author);
+		log.info("SYNC_TO_QC: Need sync: {} {}", qc, committedQC);
+
+		if (author == null) {
+			throw new IllegalStateException("Syncing required but author doesn't exist");
```
flotothemoon: "author wasn't provided"? "doesn't exist" sounds like author is non-null but we couldn't find it in some way to me

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -69,40 +71,105 @@
...
 		}
+	}
 
-		this.rootId.set(genesisVertex.getId());
-		this.vertices.put(genesisVertex.getId(), genesisVertex);
-		this.lastCommittedVertex.onNext(genesisVertex);
+	private enum SyncStage {
+		PREPARING,
+		GET_COMMITTED_VERTICES,
+		SYNC_TO_COMMIT, GET_QC_VERTICES
```
flotothemoon: should be on a new line according to our style guide
```suggestion
		SYNC_TO_COMMIT,
		GET_QC_VERTICES
```

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -115,73 +182,164 @@ public void processGetVerticesRequest(GetVerticesRequest request) {
...
+		}
+
+		syncState.setSyncStage(SyncStage.GET_COMMITTED_VERTICES);
+		syncing.put(vertexId, syncState);
+
+		log.info("SYNC_VERTICES: Committed: Sending initial GetVerticesRequest {}", syncState.getCommittedQC());
+		syncVerticesRPCSender.sendGetVerticesRequest(vertexId, syncState.author, 3, vertexId);
+	}
+
+
```
flotothemoon: extra newline here

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -194,18 +352,23 @@ private boolean addQC(QuorumCertificate qc) {
...
-				|| highestCommitted.getCommitted().get().getStateVersion() < vertexMetadata.getStateVersion()) {
+			QuorumCertificate highestCommitted = highestCommittedQC.get();
+			Optional<VertexMetadata> highest = highestCommitted.getCommitted();
+			if (highest.isPresent()) {
+				if (highest.get().getView().compareTo(vertexMetadata.getView()) < 0) {
+					this.highestCommittedQC.set(qc);
+				}
+			} else {
 				this.highestCommittedQC.set(qc);
 			}
```
flotothemoon: Could be simplified:
```suggestion
			if (!highest.isPresent() || highest.get().getView().compareTo(vertexMetadata.getView()) < 0) {
				this.highestCommittedQC.set(qc);
			}
```

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -69,41 +71,116 @@
...
-			syncedStateComputer.execute(committedGenesis);
+		for (Vertex vertex : vertices) {
+			if (!addQC(vertex.getQC())) {
+				throw new IllegalStateException(String.format("Missing qc=%s", vertex.getQC()));
+			}
+
+			try {
+				insertVertexInternal(vertex);
+			} catch (VertexInsertionException e) {
+				throw new IllegalStateException("Could not insert vertex " + vertex);
```
flotothemoon: Adding the insertion error as a cause to the exception would be nice

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/SafetyCheck.java
talekhinezh: Not sure what that gains since deterministic iteration order doesn't matter in this logic.
Also, there's already a lot of "non-determinism" in here like iterating through keys of the map or `.reduce` function above.

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -115,71 +192,155 @@ public void processGetVerticesRequest(GetVerticesRequest request) {
...
+		syncState.setSyncStage(SyncStage.GET_QC_VERTICES);
+		log.info("SYNC_VERTICES: QC: Sending initial GetVerticesRequest for qc={}", syncState.getQC());
 		syncVerticesRPCSender.sendGetVerticesRequest(vertexId, syncState.author, 1, vertexId);
 	}
 
+	private void doCommittedSync(SyncState syncState) {
+		final Hash vertexId = syncState.getQC().getProposed().getId();
+		syncState.setSyncStage(SyncStage.GET_COMMITTED_VERTICES);
+		log.info("SYNC_VERTICES: Committed: Sending initial GetVerticesRequest {}", syncState.getCommittedQC());
+		syncVerticesRPCSender.sendGetVerticesRequest(vertexId, syncState.author, 3, vertexId);
```
flotothemoon: What is the reason for the magic 3 here?

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
```diff
@@ -115,71 +192,155 @@ public void processGetVerticesRequest(GetVerticesRequest request) {
...
-			// failed
-			return;
+		long stateVersion = syncState.committedVertexMetadata.getStateVersion();
+		List<ECPublicKey> signers = Collections.singletonList(syncState.author);
+		syncState.fetched.addAll(response.getVertices());
+
+		if (syncedStateComputer.syncTo(stateVersion, signers, syncTo)) {
+			rebuildAndSyncQC(syncState);
+		} else {
+			syncState.syncStage = SyncStage.SYNC_TO_COMMIT;
```
flotothemoon: Should this be using the setter for `syncStage`?

clstrfsck (approved):

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: want the three children of committedQC so that we can construct a valid committed root

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: yes

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: cool

radixdlt/src/main/java/com/radixdlt/consensus/VertexStore.java
talekhinezh: no shouldn't be a malicious response since the RPC layer should be handling matching requests/responses. This signals that the sync was handled by a different sync and was already processed.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/checks/SafetyCheck.java
talekhinezh: Breaking it out I think will make it harder to follow since this check requires a bit of state (e.g. committedVertices map). Could break it out if there was a bigger refactor to enable moving some state to the class level but that would need to be in the future.

flotothemoon (approved):

PR #106, opened 2020-06-05
Fix for RPNV1-277 - probeTaskTest junit test failing

motoras:
I add  cleanup the code in places where resources were not released before another test was running. However we(me and Martin) are not 100% sure that the issue will never occur(while I did run the tests in an infinite loop on my machine for some time without any crash).


radixdlt/src/test/java/com/radixdlt/consensus/liveness/ScheduledTimeoutSenderTest.java
```diff
@@ -53,6 +56,11 @@ public void setUp() {
 		this.scheduledTimeoutSender = new ScheduledTimeoutSender(this.executorService);
 	}
 
+	@After
+	public void tearDown() {
+	    executor.shutdown();
```
clstrfsck: Probably should also call `awaitTermination(...)` here to ensure that all threads have stopped.

Something like: 
```
executor.shutdown();
if (!executor.awaitTermination(1, TimeUnit.SECOND)) {
    throw new IllegalStateException("Threads did not stop");
}
```

clstrfsck (changes requested):

clstrfsck (approved):

PR #108, opened 2020-06-10
Fix for bug RPNV1-260

motoras:
For details see:https://radixdlt.atlassian.net/browse/RPNV1-260
The solution simply clears the interrupted flag on the test setup method.

### DIFF
--- a/radixdlt/src/test/java/com/radixdlt/network/transport/udp/UDPTransportOutboundConnectionTest.java
+++ b/radixdlt/src/test/java/com/radixdlt/network/transport/udp/UDPTransportOutboundConnectionTest.java
@@ -81,6 +81,7 @@ public class UDPTransportOutboundConnectionTest {
 
 	@Before
 	public void setUp() throws UnknownHostException {
+	    Thread.interrupted();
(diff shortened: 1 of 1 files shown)


radixdlt/src/test/java/com/radixdlt/network/transport/udp/UDPTransportOutboundConnectionTest.java
```diff
@@ -81,6 +81,7 @@ public UDPTransportOutboundConnectionTest(String sourceAddressStr, String destin
 
 	@Before
 	public void setUp() throws UnknownHostException {
+	    Thread.interrupted();
```
flotothemoon: What does this serve?
Also, indent looks a bit weird

motoras: Indentation fixed. The interrupted flag gets cleared after calling the Thread.interrupted() method, and so will not be propagated to the test which is just about to be executed.

clstrfsck (approved):
Assuming Travis gives the all clear, this looks OK to me.

PR #114, opened 2020-06-17
Shard related code removed

motoras:
For details see [ticket](https://radixdlt.atlassian.net/browse/RPNV1-309)


radixdlt/src/main/java/org/radix/GenerateNodeKey.java
```diff
@@ -17,45 +17,36 @@
```
clstrfsck: This file should be deleted.  It was for a specific purpose to generate keys in specific shards for 1mtps project, and is no longer required.

clstrfsck (changes requested):
Please remove the commented out GenerateNodeKey class, otherwise looks good.

radixdlt/src/main/java/org/radix/GenerateNodeKey.java
motoras: Done

talekhinezh (approved):

clstrfsck (approved):
LGTM

PR #117, opened 2020-06-24
First naive go at proposal load balanaced test (deterministic)

Sajjon:
[Jira ticket](https://radixdlt.atlassian.net/browse/RPNV1-248)

### DIFF
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/BFTDeterministicTest.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/BFTDeterministicTest.java
@@ -20,6 +20,7 @@ package com.radixdlt.consensus.deterministic;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Streams;
 import com.radixdlt.consensus.deterministic.ControlledBFTNetwork.ChannelId;
 import com.radixdlt.consensus.deterministic.ControlledBFTNetwork.ControlledMessage;
 import com.radixdlt.consensus.liveness.WeightedRotatingLeaders;
@@ -31,7 +32,9 @@ import com.radixdlt.crypto.ECPublicKey;
 import com.radixdlt.identifiers.EUID;
 import com.radixdlt.utils.UInt256;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Random;
 import java.util.function.BiPredicate;
 import java.util.function.BooleanSupplier;
@@ -70,7 +73,10 @@ public class BFTDeterministicTest {
 			.map(key -> new ControlledBFTNode(
 				key,
 				network.getSender(key.getPublicKey()),
-				new WeightedRotatingLeaders(validatorSet, Comparator.comparing(v -> v.nodeKey().euid()), 5),
+				new WeightedRotatingLeaders(
+						validatorSet,
+						Comparator.comparing(v -> v.nodeKey().euid()),
+						5),
 				validatorSet,
 				enableGetVerticesRPC,
 				syncedSupplier
@@ -95,6 +101,7 @@ public class BFTDeterministicTest {
 
 	public void processNextMsg(Random random, BiPredicate<Integer, Object> filter) {
 		List<ControlledMessage> possibleMsgs = network.peekNextMessages();
+
 		if (possibleMsgs.isEmpty()) {
 			throw new IllegalStateException("No messages available (Lost Responsiveness)");
 		}
@@ -111,4 +118,9 @@ public class BFTDeterministicTest {
 	public SystemCounters getSystemCounters(int nodeIndex) {
 		return nodes.get(nodeIndex).getSystemCounters();
 	}
+
+	public ControlledBFTNetwork getNetwork() {
+		return network;
+	}
+
 }
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/BFTDeterministicTest.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/BFTDeterministicTest.java
@@ -129,9 +129,4 @@ public class BFTDeterministicTest {
 	public SystemCounters getSystemCounters(int nodeIndex) {
 		return nodes.get(nodeIndex).getSystemCounters();
 	}
-
-	public ControlledBFTNetwork getNetwork() {
-		return network;
-	}
-
 }
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/BFTDeterministicTest.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/BFTDeterministicTest.java
@@ -20,7 +20,6 @@ package com.radixdlt.consensus.deterministic;
 import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.Streams;
 import com.radixdlt.consensus.deterministic.ControlledBFTNetwork.ChannelId;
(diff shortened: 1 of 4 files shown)


radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledBFTNetwork.java
```diff
@@ -56,6 +61,14 @@
 					key -> new LinkedList<>()
 				)
 			);
+
+		this.numberOfProposalsByNodes = new HashMap<>();
```
talekhinezh: I think these proposals should be computed outside of the network as it is a property that can be measured from the outside

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/checks/ProposerLoadTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+	@Test
+	public void when_run_4_correct_nodes_with_channel_order_random_and_timeouts_disabled__then_proposal_load_should_be_completely_uniform() {
+		final Random random = new Random(12345);
+		int nodeCount = 4;
+		final BFTDeterministicTest test = new BFTDeterministicTest(nodeCount, false);
+
+		test.start();
+		for (int step = 0; step < 10_000; step++) {
+			test.processNextMsg(random);
```
talekhinezh: Should retrieve the next message here and keep track of number of proposals per node over "time".

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledBFTNetwork.java
Sajjon: more details plz :) ! How? Do you mean from `	private final ImmutableList<ControlledBFTNode> nodes;` within `BFTDeterministicTest` and then `SystemCounters systemCounters` from within `ControlledBFTNode`? If you mean the `SystemCounters` which CounterType? The `CONSENSUS_VIEW` ? Or did you mean in any other way?

Sajjon: Or did you mean not at all and then change the `processNextMsg` to return messages and check for (type cast) proposals?

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/checks/ProposerLoadTest.java
```diff
@@ -0,0 +1,65 @@
...
+		final Random random = new Random(12345);
+		int nodeCount = 4;
+		final BFTDeterministicTest test = new BFTDeterministicTest(nodeCount, false);
+
+		test.start();
+		HashMap<ECPublicKey, Integer> numberOfProposalsByNodes = new HashMap<>();
+		int stepCount = 10_000;
+		for (int step = 0; step < stepCount; step++) {
+			ControlledBFTNetwork.ControlledMessage message = test.processNextMsg(random);
+		 	if (message.getMsg() instanceof Proposal) {
```
Sajjon: @talekhinezh why did you wanna perform the diff check _**inside** this loop_, I’m using randomness for processNextMsg…. so the max diff seen so far is around 6.

But if I were to do it after the for loop, I can rely on statistical distribution and use a much smaller diff, typically 1

also kind of performance heavy to do the diff on every new proposal…?

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/checks/ProposerLoadTest.java
```diff
@@ -0,0 +1,82 @@
...
+
+				if (turn >= nodeCount) {
+					turn = 0;
+					round++;
+
+					int x = 10; // arbitrarily chosen, sorry 🤷‍♂️
+					// Check every `nodeCount*X`th (arbitrarily chosen) round
+					if (round % (nodeCount * x) == 0) {
+						System.out.println(String.format("round: %d, step: %d", round, step));
+						assertMaxValueDiff(numberOfProposalsByNodes, sender, 1);
```
talekhinezh: if this is round robin, looks like we will only ever be comparing one key to all the others which means that it's possible that there is a difference of 2 somewhere within the other nodes.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/BFTDeterministicTest.java
```diff
@@ -82,19 +85,33 @@ public void start() {
...
+	/**
+	 * Returns the next message that should be processed
+	 * @return the next message that should be processed
+	 */
+	public ControlledMessage processNextMsg(int toIndex, int fromIndex, Class<?> expectedClass) {
 		ChannelId channelId = new ChannelId(pks.get(fromIndex), pks.get(toIndex));
 		Object msg = network.popNextMessage(channelId);
 		assertThat(msg).isInstanceOf(expectedClass);
 		nodes.get(toIndex).processNext(msg);
+		return new ControlledMessage(channelId.getSender(), channelId.getReceiver(), msg);
```
talekhinezh: Would rather not use the existing ControlledMessage. Should create a different class for this tests are abstracted from ECPublicKeys and only use indices instead.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledBFTNetwork.java
```diff
@@ -96,7 +97,7 @@ public String toString() {
 		}
 	}
 
-	static final class ControlledMessage {
+	public static final class ControlledMessage {
```
talekhinezh: As mentioned above, should remove public

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/checks/ProposerLoadTest.java
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	@Test
+	public void when_run_4_correct_nodes_with_channel_order_random_and_timeouts_disabled__then_proposal_load_should_be_approx_uniform() {
+		final Random random = new Random(12345);
+		int nodeCount = 4;
+		final BFTDeterministicTest test = new BFTDeterministicTest(nodeCount, false);
+
+		test.start();
+		HashMap<ECPublicKey, Integer> numberOfProposalsByNodes = new HashMap<>();
```
talekhinezh: Change this to use indices instead of keys

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/checks/ProposerLoadTest.java
```diff
@@ -0,0 +1,82 @@
...
+		int round = 0;
+		for (int step = 0; step < 10_000; step++) {
+			ControlledBFTNetwork.ControlledMessage message = test.processNextMsg(random);
+		 	if (message.getMsg() instanceof Proposal) {
+				ECPublicKey sender = message.getChannelId().getSender();
+				numberOfProposalsByNodes.merge(sender, 1, Integer::sum);
+
+				turn++;
+
+				if (turn >= nodeCount) {
```
talekhinezh, line 63: this is still only checking just a single sender. why not just remove this check and test on every proposal?

talekhinezh: or if you want, randomise the check

Sajjon: @talekhinezh Can you please check out my branch and implement the suggestions yourself, will will notice that the test fails if you check too often :)

talekhinezh: yes will do. it means then either the test or our implementation or our understanding is broken :)

Sajjon:
Closing since stopped working with these, branch can probably be deleted...

PR #119, opened 2020-06-25
Epochs

talekhinezh:
* Refactor EpochManager to create new actors on epoch changes
* Refactor genesis creation
* Refactor VertexStore events
* Add epoch parameter to Vertex


radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
```diff
@@ -20,78 +20,91 @@
...
-		log.info("NEXT_EPOCH: ProposerElection: {}", proposerElection);
+		Pacemaker pacemaker = pacemakerFactory.create();
+		SafetyRules safetyRules = new SafetyRules(this.selfKey, SafetyState.initialState(), this.hasher);
+		PendingVotes pendingVotes = new PendingVotes(this.hasher);
+
+		VertexMetadata ancestorMetadata = epochChange.getAncestor();
+		Vertex genesisVertex = Vertex.createGenesis(ancestorMetadata);
+		QuorumCertificate genesisQC = QuorumCertificate.ofGenesis(genesisVertex);
+
+		VertexStore vertexStore = vertexStoreFactory.create(genesisVertex, genesisQC);
```
clstrfsck: Not sure why this isn't assigned directly into `this.vertexStore`?

If there is a good reason for this, the local variable *really* should have a different name than the field.

radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
```diff
@@ -105,13 +118,67 @@ public BFTEventProcessor nextEpoch(ValidatorSet validatorSet) {
 			counters
 		);
 
-		return new BFTEventPreprocessor(
+		this.currentEpoch = genesisVertex.getEpoch();
+		this.vertexStore = vertexStore;
```
clstrfsck: See comment at line 97.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledBFTNetwork.java
```diff
@@ -207,5 +208,13 @@ public void sendVote(Vote vote, ECPublicKey leader) {
 		public void committedStateSync(CommittedStateSync committedStateSync) {
 			putMesssage(new ControlledMessage(sender, sender, committedStateSync));
 		}
+
+		@Override
+		public void committedVertex(Vertex vertex) {
```
Sajjon, line 213: hmm maybe throw an exception or log something if this empty method is used? We don't want to accidentally be using methods that don't do anything? Might lead to unexpected behaviour?

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -142,7 +144,11 @@ public Builder checkNoSyncExceptions(String checkName) {
 		public Builder checkAllProposalsHaveDirectParents(String checkName) {
 			this.checksBuilder.put(checkName, new AllProposalsHaveDirectParentsCheck());
 			return this;
+		}
 
+		public Builder checkNoneCommitted(String checkName) {
+			this.checksBuilder.put(checkName, new NoneCommittedCheck());
```
Sajjon, line 150: what is a "check name"? Shoulds "checks" be renamed "milestone"? Or "threshold" or is it more like a validation/prerequisites/fulfillment? This would allow for variable names such as:

`nameOfFulfilledRequirement` or similar, which IMO is a better variable name than `checkName`?

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
```diff
@@ -160,26 +166,14 @@ public static Builder builder() {
...
-	 */
-	public Map<String, Optional<BFTCheckError>> run(long duration, TimeUnit timeUnit) {
-		TestEventCoordinatorNetwork network = TestEventCoordinatorNetwork.builder()
-			.latencyProvider(this.latencyProvider)
-			.build();
-		SimulatedBFTNetwork bftNetwork =  new SimulatedBFTNetwork(nodes, network, pacemakerTimeout, getVerticesRPCEnabled);
+	private Observable<Pair<String, Optional<BFTCheckError>>> runChecks(RunningNetwork runningNetwork, long duration, TimeUnit timeUnit) {
 		List<Pair<String, Observable<Pair<String, BFTCheckError>>>> assertions = this.checks.keySet().stream()
 			.map(name -> {
 				BFTCheck check = this.checks.get(name);
```
Sajjon, line 172: I think `BFTCheck` ought to be renamed, it is an "actor", some entity performing something, by calling invoking the method `check`. Thus `BFTChecker checker` reads out better and then `checker.check` reads out much better than `check.check`. 

But why not name it `BFTConsensusAsserter` and method `assert`, which is what the documentation of the method implies and `assert` is a much better word - since it is more specific - than `check` which does not carry the same meaning.

radixdlt/src/main/java/com/radixdlt/consensus/BasicEpochChangeRx.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Basic Epoch Change which emits an epoch with a known ancestor vertex and a list of nodes
+ */
+public final class BasicEpochChangeRx implements EpochChangeRx {
+	private final ValidatorSet validatorSet;
+	private final VertexMetadata ancestor;
+
+	public BasicEpochChangeRx(VertexMetadata ancestor, List<ECPublicKey> nodes) {
+		this.ancestor = ancestor;
+		this.validatorSet = ValidatorSet.from(
```
clstrfsck, line 37: Should the `ValidatorSet` have an epoch number associated with it as well?  If it doesn't, it is going to be difficult to be sure that QCs/Votes etc are being validated by the right validators.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -44,7 +43,6 @@
 
 	private SafetyState state;
 
-	@Inject
 	public SafetyRules(
 		@Named("self") ECKeyPair selfKey,
```
clstrfsck: Could remove Named annotation now that Inject is gone.

radixdlt/src/main/java/org/radix/api/http/RadixHttpServer.java
```diff
@@ -235,11 +235,13 @@ private void addTestRoutesTo(RoutingHandler handler) {
 		}, handler);
 
 		addGetRoute("/api/vertices/highestqc", exchange -> {
-			QuorumCertificate highestQC = this.vertexStore.getHighestQC();
-			JSONObject highestQCJson = new JSONObject();
-			highestQCJson.put("view", highestQC.getView());
-			highestQCJson.put("vertexId", highestQC.getProposed().getId());
-			respond(highestQCJson, exchange);
+			this.vertexStoreEventsRx.highQCs().firstOrError()
```
clstrfsck: Really should be handling the disposable here.

Sajjon (commented):
Grasping at straws, just wrote about naming, where I often have opinions :) feel free to ignore! It was a massive PR of which I did not understand much, so not sure how valuable this review was :)

clstrfsck (changes requested):
Couple of minor issues that need resolving, and ValidatorSet may need an epoch number.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledBFTNetwork.java
talekhinezh: in this case, calling this method is okay. it just means that we don't care about these events but I can add comments regarding this

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
talekhinezh: perhaps `invariant` is the best but would rather do this in a different PR.

radixdlt/src/main/java/com/radixdlt/consensus/BasicEpochChangeRx.java
talekhinezh: no I don't think so, that would just be an instance of replicating unnecessarily. We want epoch logic to be mostly hidden from BFTReducer.

radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
talekhinezh: yup good point

clstrfsck (approved):
LGTM

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/BFTSimulatedTest.java
flotothemoon: `assert` is a language keyword, so we can't use that

PR #120, opened 2020-06-29
Initial implementation of BFT timestamps shared in vote messages

clstrfsck:


radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -93,31 +93,38 @@ public PendingVotes(Hasher hasher) {
 	 * @return The generated QC, if any
 	 */
 	public Optional<QuorumCertificate> insertVote(Vote vote, ValidatorSet validatorSet) {
+		// Only process for valid validators
 		final ECPublicKey voteAuthor = vote.getAuthor();
-		final VoteData voteData = vote.getVoteData();
-		final Hash voteHash = this.hasher.hash(voteData);
+		if (!validatorSet.containsKey(voteAuthor)) {
```
Sajjon, line 98: ah, return early, so nice! :D

Sajjon (approved):
Fina fisken!

radixdlt/src/main/java/com/radixdlt/consensus/TimestampedVoteData.java
```diff
@@ -0,0 +1,82 @@
...
+import com.radixdlt.serialization.SerializerId2;
+import java.util.Objects;
+import javax.annotation.concurrent.Immutable;
+
+/**
+ * {@link VoteData} with a timestamp.
+ */
+@Immutable
+@SerializerId2("consensus.timestamped_vote_data")
+public final class TimestampedVoteData {
```
flotothemoon, line 36: What are the benefits of adding `nodeTimestamp` via composition rather than just by adding it to `VoteData` directly?

radixdlt/src/main/java/com/radixdlt/consensus/QuorumCertificate.java
```diff
@@ -84,10 +99,43 @@ public VoteData getVoteData() {
...
+		}
+		UInt256 median = totalPower.shiftRight(); // Divide by 2
+		// Sort ascending by timestamp
+		weightedTimes.sort(Comparator.comparing(Pair::getFirst));
+		for (Pair<Long, UInt256> w : weightedTimes) {
+			UInt256 weight = w.getSecond();
+			if (median.compareTo(weight) < 0) {
+				return w.getFirst();
+			}
+			median = median.subtract(weight);
```
flotothemoon, line 134: So we're trying to get the median time but by weight?
Wouldn't the median weight have to be averaged at "the edges" to be a proper median?

radixdlt/src/main/java/com/radixdlt/consensus/QuorumCertificate.java
```diff
@@ -84,10 +99,43 @@ public VoteData getVoteData() {
 		return voteData;
 	}
 
-	public ECDSASignatures getSignatures() {
+	public TimestampedECDSASignatures getSignatures() {
 		return signatures;
 	}
 
+	public long quorumTime(ValidatorSet validatorSet) {
```
flotothemoon: Should be `quorumTimestamp` no?

flotothemoon (approved):
Looks fine, though I have some concerns about the use of 'median'  for quorum time.

radixdlt/src/main/java/com/radixdlt/consensus/TimestampedVoteData.java
clstrfsck: Then each VoteData will (probably) be different for each node, and we will need to keep them all to validate the signatures.  This would greatly increase the quantity of data in QCs and Votes etc, particularly for circa 100 nodes.

radixdlt/src/main/java/com/radixdlt/consensus/QuorumCertificate.java
clstrfsck: Hmm.  I thought about this, but I think the timestamp really should be one of the submitted timestamps, rather than an average and the chances of the median "falling in the gap" seems pretty low.

radixdlt/src/main/java/com/radixdlt/consensus/QuorumCertificate.java
clstrfsck: Yeah, prolly.  I'll change it.

radixdlt/src/main/java/com/radixdlt/consensus/bft/VertexStore.java
```diff
@@ -492,7 +496,7 @@ public VertexMetadata insertVertex(Vertex vertex) throws VertexInsertionExceptio
 		}
 
 		for (Vertex committed : path) {
-			CommittedAtom committedAtom = new CommittedAtom(committed.getAtom(), commitMetadata);
+			CommittedAtom committedAtom = new CommittedAtom(committed.getAtom(), commitQC);
```
talekhinezh, line 499: Let's add a comment that we will want to remove QC from CommittedAtom at some point and just include data and signatures required to verify atom was committed. Otherwise QC will contain extra info (proposed and parent data) which is not required at the Ledger layer.

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
```diff
@@ -90,7 +92,7 @@ public SyncedRadixEngine(
 		StateSyncNetwork stateSyncNetwork
 	) {
 		if (epochChangeView.isGenesis()) {
-			throw new IllegalArgumentException("Epoch change view must not be genesis.");
+			Panicker.panic("Epoch change view must not be genesis.");
```
talekhinezh, line 95: Not sure if should be using Panicker within these modules as it breaks unit testing. I suggest we use panicker only from outside the modules.

radixdlt/src/main/java/com/radixdlt/utils/Panicker.java
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	/**
+	 * An error thrown by {@code Panicker.panic(...)} methods when a
+	 * panic occurs and the {@code exitOnPanic} flag is set to false.
+	 *
+	 * @see Panicker#setExitOnPanic(boolean)
+	 */
+	@VisibleForTesting
+	public static final class PanicError extends Error {
```
talekhinezh, line 63: Im a bit doubtful on use of this Error. It seems to me that any application error should at worst always be catchable.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
```diff
@@ -41,19 +43,21 @@
...
+	private final TimeSupplier timeSupplier;
 	private SafetyState state;
 
 	public SafetyRules(
 		ECKeyPair selfKey,
 		SafetyState initialState,
 		Hasher hasher,
-		HashSigner signer
+		HashSigner signer,
+		TimeSupplier timeSupplier
```
talekhinezh: Having the timeSupplier in SafetyRules seems to be the wrong place. It should only be in charge of ensuring safety and signing things and not need to retrieve time from another module. I think it should be given the time in the `voteFor` method.

radixdlt/src/main/java/com/radixdlt/consensus/TimestampedVoteData.java
talekhinezh: I'm not sure if I follow this properly. Doesn't the timestamp also need to be moved through the three chain in order to have consensus on it? Or am I missing how that's implemented?

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
clstrfsck: Panicker has support for unit testing.

radixdlt/src/main/java/com/radixdlt/utils/Panicker.java
clstrfsck: Anything that is a Throwable is catchable?  For example, see PanickerTest which tests that this is thrown.

Also, this is only here to support (unit) testing, when we don't want Panicker to actually exit — see documentation for Panicker.setExitOnPanic(...).

radixdlt/src/main/java/com/radixdlt/consensus/TimestampedVoteData.java
clstrfsck: Each Vote has a timestamp, and this is aggregated into the QC via VoteData (no timestamps) and TimestampedECDSASignatures (timestamps and signatures) once sufficient votes have been received.  All of this is verifiable by trivially reconstructing a TimestampedVoteData from the VoteData and timestamp and verifying the signature.

There is no consensus on timestamps in the sense that the protocol will drop votes/timestamps that don't conform to some ideal, otherwise this will cause a downstream problem with liveness, as discussed previously.

More details in the design https://radixdlt.atlassian.net/wiki/spaces/RPNV1/pages/1047756869/Consensus+time

radixdlt/src/main/java/com/radixdlt/consensus/bft/VertexStore.java
clstrfsck: Not sure I understand this.  A QC is just VoteData plus signatures (and now timestamps).  The VoteData is what is signed in a Vote, so that is required to verify signatures.

Could you elaborate?  Maybe I'm missing something here.

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyRules.java
clstrfsck: Yeah, that makes sense.  Will change.

radixdlt/src/main/java/com/radixdlt/consensus/TimestampedVoteData.java
talekhinezh: right but there's an issue with consensus in that everyone won't be able to agree on the same median timestamp unless it goes through the three phases

radixdlt/src/main/java/com/radixdlt/consensus/bft/VertexStore.java
talekhinezh: Yes, VoteData also contains data for the proposed and the parent vertices, data which should be opaque to the Ledger/StateComputer. If you see Libra, they use a structure called LedgerInfoWithSignatures

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
talekhinezh: Ah, gotcha. Though it still seems to me to break encapsulation? What's the win with using Panicker?

clstrfsck:
Replaced by https://github.com/radixdlt/radixdlt-core/pull/167

PR #122, opened 2020-06-30
[DONT MERGE] ☣️  security critical! Made it possible to skip signature verification, useful for tests. In…

Sajjon:
… genaral tests seem to be about 30% faster when skipping sig verification.


clstrfsck:
Going to close this, as we have another solution in this area.

PR #143, opened 2020-07-10
Infrastructure/rpn1 246 liveness 1 timeout 2

clstrfsck:
Implementation of test where all proposals in a view are dropped.

Implemented timeouts, as well as some performance improvements via handling of messages, and ability to turn off hashing/signing and sig verification when it is not required.

Re-enabled some disabled tests in the anticipation that the performance improvements will allow them to work on Travis now.

### DIFF
--- a/radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
+++ b/radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
@@ -57,13 +57,18 @@ public final class ControlledNetwork {
 		// Nothing here right now
 	}
 
-	static final class EpochAndView implements Comparable<EpochAndView> {
-		private static final Comparator<EpochAndView> COMPARATOR =
-			Comparator.comparingLong((EpochAndView eav) -> eav.epoch).thenComparing(eav -> eav.view);
+	// Message ranking.  Used to implement timeouts.
+	// Messages in a particular ranking are processed in arrival order, but
+	// timeouts in particular will be put into the next ranking to ensure
+	// that they are processed after the current ranking. The rank is changed
+	// for each node whenever a new view or epoch change is seen.
+	static final class MessageRank implements Comparable<MessageRank> {
+		private static final Comparator<MessageRank> COMPARATOR =
+			Comparator.comparingLong((MessageRank eav) -> eav.epoch).thenComparing(eav -> eav.view);
 		final long epoch;
 		final View view;
 
-		EpochAndView(long epoch, View view) {
+		MessageRank(long epoch, View view) {
 			this.epoch = epoch;
 			this.view = view;
 		}
@@ -75,15 +80,15 @@ public final class ControlledNetwork {
 
 		@Override
 		public boolean equals(Object o) {
-			if (!(o instanceof EpochAndView)) {
+			if (!(o instanceof MessageRank)) {
 				return false;
 			}
-			EpochAndView that = (EpochAndView) o;
+			MessageRank that = (MessageRank) o;
 			return this.epoch == that.epoch && Objects.equals(this.view, that.view);
 		}
 
 		@Override
-		public int compareTo(EpochAndView that) {
+		public int compareTo(MessageRank that) {
 			return COMPARATOR.compare(this, that);
 		}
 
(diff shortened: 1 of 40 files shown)


radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
```diff
@@ -242,7 +273,37 @@ public void highQC(QuorumCertificate qc) {
...
-			putMesssage(new ControlledMessage(sender, sender, committedStateSync));
+			putMessage(EARLIEST_POSSIBLE, new ControlledMessage(sender, sender, committedStateSync));
+		}
+
+		@Override
+		public void scheduleTimeout(LocalTimeout localTimeout, long milliseconds) {
+			putMessage(messageRank(localTimeout), new ControlledMessage(sender, sender, localTimeout));
+		}
+
+		private void putMessage(MessageRank eav, ControlledMessage controlledMessage) {
```
talekhinezh, line 284: Not sure if this is the correct interface moving forward but good enough for this PR

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/DeterministicTest.java
```diff
@@ -163,6 +196,22 @@ public void processNextMsg(Random random, BiPredicate<Integer, Object> filter) {
 		}
 	}
 
+	public void processNextMsgWithSenderAndReceiver(Random random, TriPredicate<Integer, Integer, Object> filter) {
```
talekhinezh, line 196: Can we add a TODO: comment specifying that we need to rethink/update interface. I don't think this interface is scalable for future adversaries.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/network/SimulationNodes.java
```diff
@@ -105,9 +109,12 @@ public SimulationNodes(
 
 	private ConsensusRunner createBFTInstance(ECKeyPair key) {
 		final Mempool mempool = new EmptyMempool();
-		final Hasher hasher = new DefaultHasher();
+		final Hasher nullHasher = o -> Hash.ZERO_HASH;
+		final Hasher defaultHasher = new DefaultHasher();
+		final HashSigner nullSigner = (k, h) -> new ECDSASignature();
+		final HashVerifier nullVerifier = (p, h, s) -> true;
```
talekhinezh, line 115: I think mocked hashers/signers are fine for Travis runs but I think we should be running real hashers and signers for Jenkins no?

Just a comment, no need to change anything for now

talekhinezh (approved):

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
clstrfsck: It's internal to the tests, so not as important here I think.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/DeterministicTest.java
clstrfsck: Yeah, I think there needs to be an interface which has access to the whole message queue, but probably need to work out what it is that we need and then implement.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/network/SimulationNodes.java
clstrfsck: I think null hashers/signers are fine until we have adversaries which sign incorrectly.  Not sure what adding hashing and signing does for the test until we have adversaries with incorrect signatures.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/DeterministicTest.java
clstrfsck: Also, not super convinced that using Random here is a great idea going forward either.

clstrfsck: OK, done.

PR #144, opened 2020-07-14
Idempotent atom sync

talekhinezh:
* Fix idempotency of atom sync in SyncedRadixEngine
* Refactor stored exceptions handling


radixdlt/src/main/java/com/radixdlt/api/StoredException.java
```diff
@@ -0,0 +1,50 @@
...
+package com.radixdlt.api;
+
+import com.radixdlt.engine.RadixEngineException;
+import com.radixdlt.middleware2.CommittedAtom;
+import java.util.Objects;
+
+/**
+ * A ledger conflict event
+ */
+public final class StoredException {
```
clstrfsck: This really needs a different name, as it isn't an Exception.  Perhaps StoreFailure?

radixdlt/src/main/java/com/radixdlt/middleware2/InternalMessagePasser.java
```diff
@@ -92,4 +109,24 @@ public void epochChange(EpochChange epochChange) {
...
+		return storedAtoms;
+	}
+
+	@Override
+	public void sendStoredException(CommittedAtom committedAtom, RadixEngineException e) {
+		storedExceptions.onNext(new StoredException(committedAtom, e));
+	}
+
+	@Override
+	public Observable<StoredException> storedExceptions() {
```
clstrfsck: Should change.  Perhaps storeFailures?

radixdlt/src/main/java/com/radixdlt/middleware2/InternalMessagePasser.java
```diff
@@ -92,4 +109,24 @@ public void epochChange(EpochChange epochChange) {
...
+		storedAtoms.onNext(new StoredAtom(committedAtom, indicies));
+	}
+
+	@Override
+	public Observable<StoredAtom> storedAtoms() {
+		return storedAtoms;
+	}
+
+	@Override
+	public void sendStoredException(CommittedAtom committedAtom, RadixEngineException e) {
```
clstrfsck: Maybe sendStoreFailure?

radixdlt/src/main/java/org/radix/GenerateUniverses.java
```diff
@@ -176,9 +174,8 @@ private Atom createGenesisAtom(byte magic, long timestamp, long planck) throws E
 		}
```
clstrfsck, line 171: Not new, but the client atom created on line 165 is never used, so could be removed.

clstrfsck (changes requested):
Looks good, but StoredException and associated methods need renaming.

FWIW, I'm not sure it's worth spending the time to improve test coverage for the API stuff, as that will be changing significantly in the future.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
```diff
@@ -49,19 +51,54 @@
...
+	ControlledNetwork() {
+		// Nothing here right now
+	}
+
+	// Message ranking.  Used to implement timeouts.
+	// Messages in a particular ranking are processed in arrival order, but
+	// timeouts in particular will be put into the next ranking to ensure
+	// that they are processed after the current ranking. The rank is changed
+	// for each node whenever a new view or epoch change is seen.
+	static final class MessageRank implements Comparable<MessageRank> {
```
Sajjon: Is `MessageRank` some known term? The naming does not do it for me :), tells me nothing reading the name. 

I guess we want it to be an implementation detail that it is using Epochs and View number? And semantically it represents some ordering?

So why not call it `MessageOrdering` or `AbstractTime`? Since in fact it does not hold any "message" it is just used to order some abstraction of "time"? Or `AbstractOrder`? :)

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
```diff
@@ -242,7 +273,37 @@ public void highQC(QuorumCertificate qc) {
...
+		@Override
+		public void scheduleTimeout(LocalTimeout localTimeout, long milliseconds) {
+			putMessage(messageRank(localTimeout), new ControlledMessage(sender, sender, localTimeout));
+		}
+
+		private void putMessage(MessageRank eav, ControlledMessage controlledMessage) {
+			ControlledNetwork.this.messageQueue.add(eav, controlledMessage);
+		}
+
+		private MessageRank messageRank(EpochChange epochChange) {
```
Sajjon: If we were to introduce a shared interface implemented by `EpochChange`, `NewView`, `Proposal`, `VertexMetaData` and `LocalTimeOut` that requires the existence of a `getEpoch()` method, then you can replace this 5 methods with the one single method? Feels neater? And might be beneficial in more places.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
```diff
@@ -67,34 +68,45 @@
 	private final ValidatorSet initialValidatorSet;
 	private final ControlledSender controlledSender;
 
+	public enum SyncAndTimeout {
+		NONE,
```
Sajjon: is `NONE` better modelled using `Optional< SyncAndTimeout >` ? That's how I would do it in Swift, but maybe it would have syntactical overhead here in Java land, in Swift it does not have any overhead, but is the encourage way of doing it.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
```diff
@@ -67,34 +68,45 @@
...
 
-
-		SyncVerticesRPCSender syncVerticesRPCSender = enableGetVerticesRPC ? sender : UnsupportedSyncVerticesRPCSender.INSTANCE;
+		SyncVerticesRPCSender syncVerticesRPCSender = (syncAndTimeout != SyncAndTimeout.NONE)
+			? sender
+			: UnsupportedSyncVerticesRPCSender.INSTANCE;
+		LocalTimeoutSender localTimeoutSender = (syncAndTimeout == SyncAndTimeout.SYNC_AND_TIMEOUT) ? sender : (v, t) -> { };
 		Mempool mempool = new EmptyMempool();
-		Hasher hasher = new DefaultHasher();
+		Hasher nullHasher = data -> Hash.ZERO_HASH;
```
Sajjon: hmm `nullHasher` is not exaclty correct though? Since we don't return `null`, but rather `ZERO_HASH`, so maybe rename it `dummyHasher`? or `alwaysZeroHasher`?

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
```diff
@@ -67,34 +68,45 @@
...
+		Hasher defaultHasher = new DefaultHasher();
+		HashSigner nullSigner = (k, h) -> new ECDSASignature();
```
Sajjon: `dummySigner`?

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
```diff
@@ -67,34 +68,45 @@
...
+		HashVerifier nullVerifier = (p, h, s) -> true;
```
Sajjon: `nullVerifier` -> `alwaysValidVerifier` ?

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/DeterministicTest.java
```diff
@@ -145,10 +163,25 @@ public void processNextMsg(int toIndex, int fromIndex, Class<?> expectedClass) {
 	}
 
 	public void processNextMsg(Random random) {
-		processNextMsg(random, (c, m) -> true);
+		processNextMsgWithReceiver(random, (c, m) -> true);
 	}
 
-	public void processNextMsg(Random random, BiPredicate<Integer, Object> filter) {
+	public void processNextMsg(Random random, Predicate<Object> filter) {
```
Sajjon: Maybe document this method to describe the behaviour of the `filter`? That returning `true` results in messages being "sent" to `nodes`?

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/DeterministicTest.java
```diff
@@ -145,10 +163,25 @@ public void processNextMsg(int toIndex, int fromIndex, Class<?> expectedClass) {
...
+		int nextIndex =  random.nextInt(possibleMsgs.size());
+		ChannelId channelId = possibleMsgs.get(nextIndex).getChannelId();
+		Object msg = network.popNextMessage(channelId);
+		int receiverIndex = pks.indexOf(channelId.getReceiver());
+		if (filter.test(msg)) {
+			nodes.get(receiverIndex).processNext(msg);
+		}
+	}
+
+	public void processNextMsgWithReceiver(Random random, BiPredicate<Integer, Object> filter) {
```
Sajjon: Document `filter`? :)

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/MessageQueue.java
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	// Really only here to work with processNextMsg(int, int, Class<?>) in BFTDeterministicTest
+	private ControlledMessage painfulPop(ChannelId channelId) {
+		List<Map.Entry<MessageRank, HashMap<ChannelId, LinkedList<ControlledMessage>>>> entries = Lists.newArrayList(this.messageQueue.entrySet());
+		Collections.sort(entries, Map.Entry.comparingByKey());
+		for (Map.Entry<MessageRank, HashMap<ChannelId, LinkedList<ControlledMessage>>> entry : entries) {
+			HashMap<ChannelId, LinkedList<ControlledMessage>> msgMap = entry.getValue();
+			LinkedList<ControlledMessage> msgs = msgMap.get(channelId);
+			if (msgs != null) {
```
Sajjon: Maybe we can avoid pyramid of doom as much as possible? This is easily done by using `continue`, like so:

Instead of:
```java
if (msgs != null) {
    // Oh no, I have to indent all my code one level!
}
```

We can do:

```java
if (msgs == null) {
    continue;
}
// Now code does not need to be indented :) 
```

radixdlt/src/main/java/com/radixdlt/api/StoredException.java
Sajjon: does not the suffix `*Exception` entail that the class is to inherit/implement Java `Exception`? This seems like an **_event_** rather? Ah which JavaDoc says, so maybe `StoreAtomConflictEvent` ?

radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
```diff
@@ -183,6 +185,33 @@ public void processEpochChange(EpochChange epochChange) {
 		queuedEvents.remove(nextEpoch);
 	}
 
+	private void logEpochChange(EpochChange epochChange, String message) {
+		if (log.isInfoEnabled()) {
```
Sajjon:
```java
if (!log.isInfoEnabled()) {
     return;
}
```

And then remove one level of indentation? :)

radixdlt/src/main/java/com/radixdlt/consensus/HashSigner.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.radixdlt.crypto.ECDSASignature;
+import com.radixdlt.crypto.ECKeyPair;
+import com.radixdlt.crypto.Hash;
+
+/**
+ * Signs a hash.
+ */
+@FunctionalInterface
+public interface HashSigner {
```
Sajjon: This looks like a duplication of the existing interface in `radixdlt-java-common` called `Signing` (which `ECKeyPair` implements by `implements Signing<ECDSASignature>`). 

So maybe rather we should rework `Signing` interface?

radixdlt/src/main/java/com/radixdlt/consensus/HashVerifier.java
```diff
@@ -15,24 +15,25 @@
...
-	}
+import com.radixdlt.crypto.ECDSASignature;
+import com.radixdlt.crypto.ECPublicKey;
+import com.radixdlt.crypto.Hash;
 
+/**
+ * Verifies signatures against hashes.
+ */
+@FunctionalInterface
+public interface HashVerifier {
```
Sajjon: Hmm you verify `signatures` not hashes, which also the JavaDoc implies...? `SignatureOfHashVerifier`? And also this should probably be moved to `radixdlt-java-common`? Should probably also be marked `SecurityCritical`? And be generalized as interface `Signing` is, and be implemented by `ECKeyPair`?

radixdlt/src/main/java/org/radix/api/observable/AtomEventObserver.java
```diff
@@ -110,14 +109,12 @@ public void start() {
...
-	public void tryNext(AtomEventWithDestinations atomEvent) {
-		if (atomEvent instanceof AtomStoredEvent) {
-			if (atomQuery.filter(atomEvent.getDestinations())) {
-				final Atom rawAtom = ClientAtom.convertToApiAtom(atomEvent.getAtom().getClientAtom());
-				final AtomEventDto atomEventDto = new AtomEventDto(AtomEventType.STORE, rawAtom);
-				synchronized (this) {
-					this.currentRunnable = currentRunnable.thenRunAsync(() -> update(atomEventDto), executorService);
-				}
+	public void tryNext(StoredAtom storedAtom) {
+		if (atomQuery.filter(storedAtom.getDestinations())) {
```
Sajjon: `if (!condition) { return; }`
and remove one level of indentation?

radixdlt/src/test/java/com/radixdlt/api/StoredExceptionTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+
+import com.radixdlt.engine.RadixEngineException;
+import com.radixdlt.middleware2.CommittedAtom;
+import org.junit.Before;
+import org.junit.Test;
+
+public class StoredExceptionTest {
```
Sajjon: what is this test really testing? it is not clear to me. Just there to increase code coverage?

Sajjon (changes requested):
Only big remarks is `HashVerififer` and `HashSigner` which needs to be moved I think and potentially remove `HashSigner` and reuse `Signing`? what do you think? :)

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
talekhinezh: Not sure about the confusion here. It simply is a class which represents the rank of a message. `MessageOrdering` would imply more of a `Comparator`. Anyway, this class isn't actually a part of this PR.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
talekhinezh: Also not part of this PR.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/DeterministicTest.java
talekhinezh: Also not part of PR.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/DeterministicTest.java
talekhinezh: Also not part of PR.

radixdlt/src/main/java/com/radixdlt/consensus/EpochManager.java
talekhinezh: Also not part of PR.

radixdlt/src/main/java/com/radixdlt/consensus/HashSigner.java
talekhinezh: Not part of PR.

radixdlt/src/main/java/com/radixdlt/consensus/HashVerifier.java
talekhinezh: Not part of PR

radixdlt/src/test/java/com/radixdlt/api/StoredExceptionTest.java
talekhinezh: yes, for the most part

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
talekhinezh: Not part of PR.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
talekhinezh: Not part of PR.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
talekhinezh: Not part of PR.

radixdlt/src/main/java/org/radix/api/jsonrpc/SubmitAtomAndSubscribeEpic.java
```diff
@@ -95,32 +94,36 @@ public void onStored(boolean first) {
...
-					ParticleConflictException particleConflictException = (ParticleConflictException) e;
-					ParticleConflict conflict = particleConflictException.getConflict();
-					JSONObject data = new JSONObject();
-					data.put("pointerToIssue", conflict.getDataPointer().toString());
-					data.put("message",
-							conflict.getAtomIds().stream().filter(a -> !a.equals(atomId)).findAny().map(Object::toString).orElse(null));
+			public void onStoredFailure(StoredFailure e) {
+				RadixEngineException exception = e.getException();
+				JSONObject data = new JSONObject();
+				data.put("pointerToIssue", exception.getDataPointer().toString());
```
Sajjon, line 100: We ought to avoid using string literals for JSON keys... change into model object? :)

Sajjon (commented):
OK so mostly commented on incorrect code :D hmm anyway, I think we should not be using string literals for JSON keys, maybe change to POJOs?

radixdlt/src/main/java/org/radix/api/jsonrpc/SubmitAtomAndSubscribeEpic.java
talekhinezh: no not worth it at the moment

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
Sajjon: > MessageOrdering would imply more of a Comparator
Well yes, that is basically the only thing that the class does right now, with code:

```java
private static final Comparator<MessageRank> COMPARATOR =
			Comparator.comparingLong((MessageRank eav) -> eav.epoch).thenComparingLong(eav -> eav.view);
```

Right? :)

I'm saying that "Rank" (which really should be "Ranking" ) and that "Message" is confusing for me. It does not contain any message, and it can be used for anything really, not only messages. And ranking implies some more sophisticated ranking logic, e.g. ELO Ranking, but we only really compare Epoch+View and is only really used for **_ordering_** ??

Even if not part of PR, I think the naming still can be improved  IMO :)

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
```diff
@@ -224,7 +213,10 @@ private void handleRadixEngineException(CommittedAtom atom, RadixEngineException
 	public void execute(CommittedAtom atom) {
 		// TODO: remove lock
 		synchronized (lock) {
-			// TODO: need to implement idempotency
+			if (atom.getVertexMetadata().getStateVersion() != 0
```
flotothemoon, line 216: Non-obvious to me here why `atom.getVertexMetadata().getStateVersion() != 0` must hold

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
```diff
@@ -234,16 +226,23 @@ public void execute(CommittedAtom atom) {
...
+				this.lastStoredAtom.onNext(atom);
+				this.mempool.removeCommittedAtom(atom.getAID());
 			} else if (atom.getVertexMetadata().isEndOfEpoch()) {
 				// TODO: HACK
 				// TODO: Remove and move epoch change logic into RadixEngine
-				if (emptyCommittedAtoms.isEmpty()
-					|| emptyCommittedAtoms.getLast().getVertexMetadata().getStateVersion() != atom.getVertexMetadata().getStateVersion()) {
-					emptyCommittedAtoms.add(atom);
-				}
+				this.unstoredCommittedAtoms.add(atom);
```
flotothemoon, line 244: What are we doing with atoms put into `unstoredCommittedAtoms`?

flotothemoon: Do they ever get stored in this node? I only see them being used for sync responses

flotothemoon (approved):
Apart from everything already mentioned (which has been addressed), this looks good to merge.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
clstrfsck: It represents the rank of a message.  Not sure how better to communicate this?  A ranking is an ordering of a set of items by their rank, and this isn't one of those.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
clstrfsck: Optional doesn't cover the three states required in a consistent way IMO.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNetwork.java
clstrfsck: We could, but not sure that it makes sense to clutter up the production code to make the test code simpler.

radixdlt/src/integration/java/com/radixdlt/consensus/deterministic/ControlledNode.java
clstrfsck: Think "null" in the sense that it doesn't do any hashing.

radixdlt/src/main/java/com/radixdlt/consensus/sync/SyncedRadixEngine.java
talekhinezh: No, they need to at some point.

clstrfsck (approved):

PR #148, opened 2020-07-16
Decouple ECPublicKey from Consensus

talekhinezh:
* Refactor consensus to decouple ECPublicKey as an identifier to nodes
* Move classes to correct packages


Sajjon (approved):
Skimmed through it, LGTM

radixdlt/src/main/java/com/radixdlt/consensus/bft/BFTNode.java
```diff
@@ -0,0 +1,64 @@
...
+	public boolean equals(Object o) {
+		if (!(o instanceof BFTNode)) {
+			return  false;
+		}
+
+		BFTNode bftNodeId = (BFTNode) o;
+		return Objects.equals(bftNodeId.key, this.key);
+	}
+
+	public String getSimpleName() {
```
clstrfsck, line 61: Should this just be toString()?

radixdlt/src/main/java/com/radixdlt/consensus/HashSigner.java
```diff
@@ -30,19 +29,17 @@
 	 * Sign the specified hash with the specified key.
 	 *
 	 * @param hash The hash to sign
-	 * @param key The key to sign with
 	 * @return The {@link ECDSASignature}
 	 */
-	ECDSASignature sign(ECKeyPair key, byte[] hash);
+	ECDSASignature sign(byte[] hash);
```
clstrfsck, line 37: Not sure this is a great idea.  Everywhere we sign things, we also need the public key, and now instances of this are required to have an embedded private key which makes it harder to keep track of where private keys are used.

If we are going to go this way, this interface and implementors of it should be marked SecurityCritical(SecurityKind.SIG_SIGN) as they will all have embedded private keys.  Also should consider making the related change to HashVerifier for consistency.

radixdlt/src/main/java/org/radix/GlobalInjector.java
```diff
@@ -158,6 +160,20 @@ public RadixAddress get() {
 		}
 	}
 
+	static class SelfBFTNodeProvider implements Provider<BFTNode> {
```
clstrfsck, line 163: I think this could be converted to a provider method, given that is what we are using elsewhere?

clstrfsck (commented):
Not sure about the embedded keys in HashSigner.  Can we discuss?

PR #165, opened 2020-07-31
Implement a different mechanism for catching bft events

talekhinezh:


radixdlt/src/main/java/org/radix/universe/system/LocalSystem.java
```diff
@@ -46,28 +50,58 @@
...
-	LocalSystem(Supplier<Map<String, Object>> counters) {
+	LocalSystem(InMemoryEpochInfo epochInfo, Supplier<Map<String, Object>> counters) {
+		this.epochInfo = epochInfo;
 		this.keyPair = ECKeyPair.generateNew();
 		this.counters = counters;
 	}
 
-	public LocalSystem(Supplier<Map<String, Object>> counters, ECKeyPair key, String agent, int agentVersion, int protocolVersion, ImmutableList<TransportInfo> supportedTransports) {
+	public LocalSystem(
+		InMemoryEpochInfo epochInfo,
```
clstrfsck: I really don't like this interface, particularly the strong coupling between InMemoryEpochInfo and LocalSystem.

Can we abstract out the supply of the query details, perhaps through a generalisation of the approach for counters?

clstrfsck (changes requested):
In general looks good, but I am a bit concerned about the coupling with InMemoryEpochInfo and LocalSystem.  I think this is problematic as the coupling is tight, and it is also not something that is extensible/scalable to other bits of information that should go into node system API queries.

Can we look at generalising an interface between information suppliers and LocalSystem?  Ideally Counters should also be included in this generalisation.

Perhaps some sort of list of InformationSupplier interfaces could be used.  The InformationSupplier interface would have the necessary methods to call to extract the information to be included in the API response.

PR #166, opened 2020-08-01
Infrastructure/cleanup debug

talekhinezh:


radixdlt/src/test/java/org/radix/serialization/RadixTest.java
```diff
@@ -17,6 +17,7 @@
 
 package org.radix.serialization;
 
+import com.google.common.collect.ImmutableMap;
 import com.radixdlt.DefaultSerialization;
 import com.radixdlt.counters.SystemCounters;
```
clstrfsck: SystemCounters import not used anymore.

radixdlt/src/main/java/com/radixdlt/api/InMemoryInfoStateManager.java
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	public QuorumCertificate getHighestQC() {
+		return this.highQC.get();
+	}
+
+	public List<Vertex> getCommittedVertices() {
+		List<Vertex> vertices = Lists.newArrayList();
+		// Use internal iteration for thread safety
+		vertices.addAll(vertexRingBuffer);
```
clstrfsck: There's no guarantee here that `addAll(...)` doesn't use an iterator over `vertexRingBuffer` in its implementation, and it is `vertexRingBuffer` that needs protecting from multiple thread access.

Better would be `vertexRingBuffer.forEach(vertices::add)` which uses internal iteration from the (synchronized) vertexRingBuffer

radixdlt/src/main/java/org/radix/GlobalInjector.java
```diff
@@ -192,7 +201,35 @@ public LocalSystemProvider(SystemCounters counters, RuntimeProperties properties
 		public LocalSystem get() {
 			String host = this.hostIp.hostIp()
 				.orElseThrow(() -> new IllegalStateException("Unable to determine host IP"));
-			return LocalSystem.create(this.counters, this.properties, this.universe, host);
+			Supplier<Map<String, Object>> infoSupplier = () -> {
```
clstrfsck: This really needs to go into a separate class, or at the very least a separate method.

radixdlt/src/main/java/org/radix/universe/system/LocalSystem.java
```diff
@@ -83,10 +89,10 @@ int getJsonProcessors() {
 		return Runtime.getRuntime().availableProcessors();
 	}
 
-	public static LocalSystem create(SystemCounters counters, RuntimeProperties properties, Universe universe, String host) {
+	public static LocalSystem create(Supplier<Map<String, Object>> infoSupplier, RuntimeProperties properties, Universe universe, String host) {
```
clstrfsck: It would be much better, from a documentation point of view, if we used a specific interface here (maybe `InfoSupplier`), rather than the generic interface `Supplier<Map<String, Object>>`.  As things are, it is impossible to determine what structure the map here should have, and the structure itself is kind of non-trivial.

The only way to determine what should be passed in here as an argument is to look up the callers, and inspect the non-trivial code there, which is more work than it should be.

A simple interface with some documentation should fix this easily.  Something like:
```
/**
 * Information supplier for debug information etc etc
 */
@ FunctionalInterface
public interface InfoSupplier {
    /**
     * Returns current values for debug information arranged etc and so forth...
     */
    Map<String, Object> getInfo();
}
```

clstrfsck (changes requested):
Would like to see the interface between information suppliers and LocalSystem documented a bit better, perhaps through an interface as described.

radixdlt/src/main/java/org/radix/universe/system/LocalSystem.java
talekhinezh: True, will do.

The reason why I keep these interfaces more "open-ended" is that I don't believe these are the correct interfaces we will use in the long run (need more info/design) and imo better to have clearly incorrect interfaces which are easily reversible over building out incorrect interfaces that ends up needing a bigger refactor.

Of course, where that line is I guess is a matter of opinion so will fix this

clstrfsck (approved):

PR #173, opened 2020-08-05
Infrastructure/refactor modules for simulation

talekhinezh:


radixdlt/src/integration/java/com/radixdlt/consensus/simulation/NullCryptoModule.java
```diff
@@ -15,20 +15,23 @@
...
+import com.radixdlt.consensus.HashSigner;
+import com.radixdlt.consensus.HashVerifier;
+import com.radixdlt.consensus.Hasher;
+import com.radixdlt.crypto.ECDSASignature;
+import com.radixdlt.crypto.Hash;
 
+/**
+ * For testing where verification and signing is skipped
+ */
+public class NullCryptoModule extends AbstractModule {
```
clstrfsck, line 30: Noice.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/network/SimulationNodes.java
```diff
@@ -18,59 +18,36 @@
...
 /**
  * A multi-node bft test network where the network and latencies of each message is simulated.
  */
 public class SimulationNodes {
 	private final int pacemakerTimeout;
 	private final SimulationNetwork underlyingNetwork;
-	private final ImmutableMap<BFTNode, SystemCounters> counters;
-	private final ImmutableMap<BFTNode, InternalMessagePasser> internalMessages;
-	private final ImmutableList<ConsensusRunner> runners;
+	private final ImmutableList<Injector> nodeInstances;
```
clstrfsck, line 50: I think it is not a great idea here to use an Injector as a generic factory.  In my experience, this always becomes problematic.

Also problematic in my opinion to be creating an injector per node — Ideally we would be using the injector to set up the whole test, but I think this is going to be a lot of work, given the way we are using singletons extensively at the moment.

For now, we could create the unstarted objects at initialisation time to avoid this.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/network/SimulationNodes.java
```diff
@@ -196,7 +118,8 @@ private ConsensusRunner createBFTInstance(BFTNode self) {
 
 			@Override
 			public InfoRx getInfo(BFTNode node) {
-				return internalMessages.get(node);
+				int index = nodes.indexOf(node);
```
clstrfsck, line 121: This seems super inefficient, especially given that this turns many of its uses into an n^2 thing.  Maybe making nodes a LinkedHashMap<BFTNode, Integer> would be better?  getNodes() above could change to return a Collection<BFTNode> and return nodes.keySet() this way.

radixdlt/src/main/java/com/radixdlt/SystemInfoModule.java
```diff
@@ -0,0 +1,100 @@
...
+		return new InMemorySystemInfoManager(infoRx, vertexBufferSize, vertexUpdateFrequency);
+	}
+
+	@Provides
+	@Singleton
+	private InfoSupplier infoSupplier(
+		SystemCounters counters,
+		InMemorySystemInfoManager infoStateManager
+	) {
+		return () -> {
```
clstrfsck, line 71: I think at some point we should change this to use a guice multibinding, a bit like the way the networking stuff does multibindings for transport.  Not sure if it is really required now, but should be done at some point.

radixdlt/src/main/java/com/radixdlt/consensus/bft/VertexStore.java
```diff
@@ -121,7 +125,7 @@ public VertexStore(
 			rootQC,
 			Collections.emptyList(),
 			syncedStateComputer,
-			syncVerticesRPCSender,
+			syncVerticesRPCSender, syncedVertexSender,
```
clstrfsck, line 128: Slightly out of context formatting here.

radixdlt/src/main/java/com/radixdlt/syncer/SyncManager.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.function.Consumer;
+import java.util.function.LongConsumer;
+import java.util.function.LongSupplier;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import com.radixdlt.middleware2.CommittedAtom;
+import com.radixdlt.utils.ThreadFactories;
+
+public class SyncManager {
```
clstrfsck, line 37: Looks like this should be final?  Doesn't look extensible.

radixdlt/src/main/java/com/radixdlt/syncer/SyncManager.java
```diff
@@ -0,0 +1,185 @@
...
+		}
+		timeoutChecker = executorService.schedule(() -> sendSyncRequests(requestSender), patience, TimeUnit.SECONDS);
+	}
+
+	@VisibleForTesting
+	long getTargetVersion() {
+		return targetVersion;
+	}
+
+	void setTargetListener(LongConsumer onTarget) {
```
clstrfsck, line 164: Looks like this method and setVersionListener below should be annotated VisibleForTesting too.

radixdlt/src/test/java/com/radixdlt/CryptoModuleTest.java
```diff
@@ -0,0 +1,48 @@
...
+				@Override
+				protected void configure() {
+					bind(ECKeyPair.class).annotatedWith(Names.named("self")).toInstance(mock(ECKeyPair.class));
+				}
+			},
+			new CryptoModule()
+		);
+
+		Hasher hasher = injector.getInstance(Hasher.class);
+		assertThat(hasher).isNotNull();
```
clstrfsck, line 46: Probably should check other interfaces here too.

radixdlt/src/test/java/com/radixdlt/utils/TwoSenderToRxTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static org.mockito.Mockito.mock;
+
+import io.reactivex.rxjava3.observers.TestObserver;
+import org.junit.Test;
+
+public class TwoSenderToRxTest {
+	@Test
+	public void when_send__sync_event__then_should_receive_it() {
+		TwoSenderToRx<Object, Object, Object> senderToRx = new TwoSenderToRx<>((o1, o2) -> o1);
```
clstrfsck, line 28: Not sure this test really fully tests the interface.  Perhaps a test where the two arguments are integers and the result is the sum of them would test this better?  At the moment it would be possible for TwoSenderToRx to completely ignore the second argument to send, and still pass this test.

radixdlt/src/test/java/com/radixdlt/utils/SenderToRxTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import static org.mockito.Mockito.mock;
+
+import io.reactivex.rxjava3.observers.TestObserver;
+import org.junit.Test;
+
+public class SenderToRxTest {
+	@Test
+	public void when_send__sync_event__then_should_receive_it() {
+		SenderToRx<Object, Object> senderToRx = new SenderToRx<>(o -> o);
```
clstrfsck, line 28: Would like to see a function here that actually changes the value to ensure that the mapper is applied correctly.  An integer and increment would be fine.

radixdlt/src/integration/java/com/radixdlt/consensus/simulation/network/SimulationNodes.java
clstrfsck: Maybe if the injector thing is tidied up a bit, then `nodes` and `nodeInstances` could be combined into an ImmutableBiMap<Integer, RunnableBFTNode>, where RunnableBFTNode is a class that has a BFTNode, ConsensusRunner and InfoRx contained within.

radixdlt/src/main/java/com/radixdlt/utils/SenderToRx.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.radixdlt.utils;
+
+import io.reactivex.rxjava3.core.Observable;
+import io.reactivex.rxjava3.subjects.BehaviorSubject;
+import io.reactivex.rxjava3.subjects.Subject;
+import java.util.Objects;
+import java.util.function.Function;
+
+public final class SenderToRx<T, U> {
```
clstrfsck, line 26: Documentation.

radixdlt/src/main/java/com/radixdlt/utils/TwoSenderToRx.java
```diff
@@ -0,0 +1,41 @@
...
+
+package com.radixdlt.utils;
+
+import io.reactivex.rxjava3.core.Observable;
+import io.reactivex.rxjava3.subjects.BehaviorSubject;
+import io.reactivex.rxjava3.subjects.Subject;
+import java.util.Objects;
+import java.util.function.BiFunction;
+
+public final class TwoSenderToRx<T, U, V> {
```
clstrfsck, line 26: Documentation.

radixdlt/src/main/java/com/radixdlt/SystemInfoMessagesModule.java
```diff
@@ -0,0 +1,94 @@
...
+import io.reactivex.rxjava3.core.Observable;
+
+/**
+ * Module which routes system info related messages
+ */
+public final class SystemInfoMessagesModule extends AbstractModule {
+
+	@Override
+	protected void configure() {
+		SenderToRx<Vertex, Vertex> committedVertices = new SenderToRx<>(i -> i);
```
clstrfsck, line 38: This seems like the sort of thing that should have a static factory method.

radixdlt/src/main/java/com/radixdlt/SystemInfoMessagesModule.java
```diff
@@ -0,0 +1,94 @@
...
+		SenderToRx<QuorumCertificate, QuorumCertificate> highQCs = new SenderToRx<>(i -> i);
+		SenderToRx<Timeout, Timeout> timeouts = new SenderToRx<>(i -> i);
+		SenderToRx<EpochView, EpochView> currentViews = new SenderToRx<>(i -> i);
+
+		EpochInfoSender epochInfoSender = new EpochInfoSender() {
```
clstrfsck, line 43: There's too much creation of objects in here for a DI configuration.  Maybe doesn't need to be addressed immediately here, but something we need to consider as we change this.

clstrfsck (approved):
Generally this is pretty good.  Would like to see some changes made so we are not using guice injectors as run-time object factories, but this is not critical right now.

Some other minor suggestions, not affecting functionality.

PR #189, opened 2020-08-26
Make timestamps in Peer thread safe

clstrfsck:
Fixes an issue in Peer where timestamps are not threadsafe.

Note: Needs to be merged to RC, not infrastructure/refactor-ledger, but targeting that branch for review.

### DIFF
--- a/radixdlt/src/main/java/com/radixdlt/network/addressbook/Peer.java
+++ b/radixdlt/src/main/java/com/radixdlt/network/addressbook/Peer.java
@@ -17,9 +17,10 @@
 
 package com.radixdlt.network.addressbook;
 
-import java.util.HashMap;
+import java.util.EnumMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Stream;
 
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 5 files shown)


radixdlt/src/main/java/com/radixdlt/network/addressbook/Peer.java
```diff
@@ -46,20 +47,28 @@
...
 	public static final int DEFAULT_BANTIME = 60 * 60;
 
+	private static final long DEFAULT_TIMESTAMP = 0L;
+
 	@JsonProperty("ban_reason")
 	@DsonOutput(Output.PERSIST)
 	private String banReason;
 
-	private HashMap<String, Long> timestamps;
+	private EnumMap<Timestamps, AtomicLong> timestamps;
```
talekhinezh: Should be final as well?

talekhinezh: Also, should probably be ImmutableMap? Otherwise I think there are still thread issues between constructor and getter methods.

talekhinezh (changes requested):

radixdlt/src/main/java/com/radixdlt/network/addressbook/Peer.java
clstrfsck: Will make it final, but no threading issues, as the structure of the map doesn't change, only the values of the AtomicLongs are changed.

talekhinezh (approved):

PR #194, opened 2020-08-31
Infrastructure/move validator set into re

talekhinezh:


radixdlt/src/main/java/com/radixdlt/middleware2/store/CommittedAtomsStore.java
```diff
@@ -130,7 +132,43 @@ public void storeAtom(CommittedAtom committedAtom) {
...
+		final String idForClass = serialization.getIdForClass(particleClass);
+		final EUID numericClassId = SerializationUtils.stringToNumericID(idForClass);
+		final byte[] indexableBytes = EngineAtomIndices.toByteArray(IndexType.PARTICLE_CLASS, numericClassId);
+		SearchCursor cursor = store.search(LedgerIndexType.DUPLICATE, new StoreIndex(indexableBytes), LedgerSearchMode.EXACT);
+
+		V v = initial;
+		while (cursor != null) {
+			AID aid = cursor.get();
+			Optional<LedgerEntry> ledgerEntry = store.get(aid);
+			if (ledgerEntry.isPresent()) {
```
clstrfsck, line 151: ~~Should use ifPresent(...) here.~~

Actually can't do that, because `v` is not effectively final.

Shouldn't this be a hard fail if ledgerEntry is not present?

radixdlt/src/main/java/com/radixdlt/middleware2/store/CommittedAtomsStore.java
```diff
@@ -140,11 +178,6 @@ public void storeAtom(CommittedAtom committedAtom) {
 				.collect(ImmutableList.toImmutableList());
 	}
 
-	@Override
```
clstrfsck, line 143: Woohoo!

radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineValidatorSetBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+	private final LinkedList<ECPublicKey> lastRemovedValidators = new LinkedList<>();
+
+	public RadixEngineValidatorSetBuilder(BFTValidatorSet validatorSet) {
+		this.validators = validatorSet.getValidators().stream().map(v -> v.getNode().getKey()).collect(Collectors.toCollection(HashSet::new));
+	}
+
+	public RadixEngineValidatorSetBuilder removeValidator(RadixAddress validatorAddress) {
+		this.validators.remove(validatorAddress.getPublicKey());
+		if (lastRemovedValidators.size() == 2) {
+			lastRemovedValidators.removeFirst();
```
clstrfsck: This special magic really needs some comments to describe what is going on with lastRemovedValidators.  I can't work it out from here.

Seems to me to suggest that deregistered validators will be co-opted into being validators if there are less than two validators.  This doesn't seem reasonable.  Perhaps we should consider this a network fail?

radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineValidatorSetBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Helper class to build validator sets from radix engine state
+ */
+public final class RadixEngineValidatorSetBuilder {
+	private final HashSet<ECPublicKey> validators;
```
clstrfsck, line 37: A LinkedHashSet would preserve ordering, which may be useful.

clstrfsck (commented):
Looks OK, but I think we need some comments explaining why validators should be expected to be validators, even if they have decided to remove themselves from the pool.

radixdlt/src/main/java/org/radix/api/jsonrpc/AtomStatusEpic.java
```diff
@@ -18,11 +18,11 @@
 package org.radix.api.jsonrpc;
 
 import com.radixdlt.consensus.VertexMetadata;
-import com.radixdlt.engine.AtomStatus;
 import com.radixdlt.engine.RadixEngineException;
 import com.radixdlt.mempool.MempoolDuplicateException;
 import com.radixdlt.mempool.MempoolFullException;
 import com.radixdlt.identifiers.AID;
+import com.radixdlt.middleware2.LedgerAtom;
```
clstrfsck: This doesn't look like it's required any more.

clstrfsck (approved):

PR #203, opened 2020-09-10
Infrastructure/sync test

talekhinezh:


radixdlt/src/integration/java/com/radixdlt/integration/distributed/deterministic/DeterministicRandomlySyncedLedgerModule.java
```diff
@@ -42,19 +42,20 @@ public DeterministicRandomlySyncedLedgerModule(Random random) {
...
+	@Provides
 	@Singleton
-	@ProvidesIntoSet
 	Ledger syncedExecutor(CommittedStateSyncSender committedStateSyncSender) {
 		return new Ledger() {
 			@Override
 			public LedgerHeader prepare(VerifiedVertex vertex) {
 				return LedgerHeader.create(
-					0,
+					vertex.getParentHeader().getLedgerHeader().getEpoch(),
```
siy, line 58: Perhaps it worth to create specialized versions of LedgerHeader.create() which will cover frequent use cases like creating header from vertex and creating header from proof and validator set. Such methods will serve two purposes: reduce repetitive boilerplate and make code more readable as method name can provide more information than generic 'create'.

radixdlt/src/main/java/org/radix/Radix.java
```diff
@@ -18,8 +18,11 @@
 package org.radix;
 
 import com.google.common.collect.ImmutableMap;
-import com.radixdlt.ConsensusRunner;
+import com.google.inject.Key;
+import com.google.inject.TypeLiteral;
+import com.radixdlt.ModuleRunner;
 import com.radixdlt.DefaultSerialization;
+import com.radixdlt.consensus.VerifiedLedgerHeaderAndProof;
```
clstrfsck, line 25: Not sure this is used?

radixdlt/src/main/java/org/radix/Radix.java
```diff
@@ -18,8 +18,11 @@
...
 import com.radixdlt.statecomputer.ClientAtomToBinaryConverter;
 import com.radixdlt.sync.SyncServiceRunner;
```
clstrfsck, line 27: Not sure this is used?

radixdlt/src/main/java/org/radix/Radix.java
```diff
@@ -37,6 +40,8 @@
 import com.radixdlt.universe.Universe;
 import com.radixdlt.utils.Bytes;
 
+import java.util.Comparator;
```
clstrfsck, line 43: Not sure this is used?

clstrfsck (approved):
Looks good - minor comments about unused imports, and testing could use a bit of a boost.

radixdlt/src/integration/java/com/radixdlt/integration/distributed/deterministic/DeterministicRandomlySyncedLedgerModule.java
talekhinezh: I agree with the sentiment in certain cases. But in this case, LedgerHeader and VerifiedVertex are data classes of two different modules (consensus + ledger) and should not "know" about each other. Ideally they would never even be used together in the same class. The idea with the current architecture plan is to continually separate these two interfaces which may mean there are some "ugly" code but I think it's fine as long as architecture is correct.

PR #213, opened 2020-09-21
Updated Dockerfile.core to use jdk 11

shambupujar:


siy (approved):
LGTM

clstrfsck (approved):
I think at some point we should consider moving the common/standard parts of the yml files into the Dockerfile, but this looks fine for now.

PR #219, opened 2020-09-28
Remove reflective access to JCE classes (not necessary for Java 11)

siy:


clstrfsck (approved):
Looks good for this repo, but I think there are changes required in radix-engine-library, radixdlt-java and radixdlt-java-common too.

siy:
Seems radixdlt-java does not have references to JceSecurity class, so only 2 other repos need to be updated.

PR #225, opened 2020-09-30
Switch to factory methods

siy:


siy:
Unable to reset Travis cache

PR #228, opened 2020-10-06
Implemented SyncInfo

clstrfsck:


siy (approved):
LGTM

radixdlt/src/main/java/com/radixdlt/consensus/SyncInfo.java
```diff
@@ -0,0 +1,111 @@
...
+import com.radixdlt.serialization.SerializerDummy;
+import com.radixdlt.serialization.SerializerId2;
+import com.radixdlt.serialization.DsonOutput.Output;
+
+/**
+ * Current state of synchronisation for sending node.
+ */
+@Immutable
+@SerializerId2("consensus.sync_info")
+public final class SyncInfo {
```
talekhinezh: Can we name this something like "BFTHighQC" as imo the main purpose of this data structure is for propagating the highQC/parent for consensus. Also can we add comment highestCommittedQC->committed needs to be an ancestor of highQC->proposed. But highestCommittedQC->proposed does not need to be an ancestor of highQC->proposed.

radixdlt/src/main/java/com/radixdlt/consensus/SyncInfo.java
```diff
@@ -0,0 +1,111 @@
...
+		@JsonProperty("highest_qc") QuorumCertificate highestQC,
+		@JsonProperty("committed_qc") QuorumCertificate highestCommittedQC
+	) {
+		return new SyncInfo(highestQC, highestCommittedQC);
+	}
+
+	private SyncInfo(QuorumCertificate highestQC, QuorumCertificate highestCommittedQC) {
+		this.highestQC = Objects.requireNonNull(highestQC);
+		// Don't include separate committedQC if it is the same as highQC
+		if (highestCommittedQC == null || highestQC.equals(highestCommittedQC)) {
```
talekhinezh: Is this optimization only for serialization size? If so, should add comment.

radixdlt/src/main/java/com/radixdlt/consensus/SyncInfo.java
clstrfsck: OK, done.  Change the name to HighQC, as the BFT prefix doesn't seem apt here.

Added comments to factory method as suggested.

radixdlt/src/main/java/com/radixdlt/consensus/SyncInfo.java
clstrfsck: OK, comment added.

talekhinezh (approved):

PR #229, opened 2020-10-07
Fix Genesis Child timestamp

talekhinezh:
* Since Genesis QC is mocked it doesn't contain signed timestamps like other QCs do, this leads to the child of genesis having no timestamp.
* Fix this so that the genesis child will reuse the previous timestamp


clstrfsck (approved):
Looks good.

siy (approved):
LGTM

PR #230, opened 2020-10-08
Use Guava's HashCode and reorganize hashing utils

LukasGasior1:
Related to: https://github.com/radixdlt/radixdlt-java-common/pull/26 and https://github.com/radixdlt/radix-engine-library/pull/49

* Domain objects no longer have the `getHash` method, instead hash is calculated where needed using the injected `Hasher` implementation. As a result hashCode&equals needed to be added (as they previously relied on precalculated hash).


radixdlt/src/main/resources/default.config
```diff
@@ -240,7 +240,7 @@ network.tcp.priority=0
 
 # Directly specifies the public IP address of this node.
 # Default: Empty
-# host.ip=
+host.ip=127.0.0.1
```
siy: Looks like a leftover from local testing.

siy (approved):
LGTM

radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/application/RadixEngineValidatorRegistratorAndUnregistrator.java
```diff
@@ -44,6 +46,9 @@
  * Randomly registers and unregisters nodes as validators
  */
 public class RadixEngineValidatorRegistratorAndUnregistrator implements CommandGenerator {
+
+	private final Hasher hasher = Sha256Hasher.withDefaultSerialization();
```
clstrfsck: Can we not pass hasher into constructor?

radixdlt/src/main/java/com/radixdlt/CryptoModule.java
```diff
@@ -43,17 +42,7 @@ protected void configure() {
...
-			public Hash hash(Object o) {
-				return Hash.of(serialization.toDson(o, Output.HASH));
-			}
-
-			@Override
-			public Hash hashBytes(byte[] bytes) {
-				return Hash.of(bytes);
-			}
-		};
+		return new Sha256Hasher(serialization);
```
clstrfsck, line 45: I think @talekhinezh has made similar changes here in a different PR.

radixdlt/src/main/java/com/radixdlt/SyncMempoolServiceModule.java
```diff
@@ -51,23 +52,25 @@ SubmissionControl submissionControl(
...
 	}
 
 	@Provides
 	@Singleton
-	private AtomToClientAtomConverter converter() {
+	private AtomToClientAtomConverter converter(Hasher hasher) {
 		return atom -> {
 			try {
-				return ClientAtom.convertFromApiAtom(atom);
+				return ClientAtom.convertFromApiAtom(atom, hasher);
```
clstrfsck: This is better.

radixdlt/src/main/java/com/radixdlt/network/messaging/MessageCentralImpl.java
```diff
@@ -202,7 +204,10 @@ private void outboundMessageProcessor(MessageEvent outbound) {
 	private Message deserialize(byte[] in) {
 		try {
 			byte[] uncompressed = Snappy.uncompress(in);
+
```
clstrfsck, line 207: Excessive whitespace here.

radixdlt/src/main/java/org/radix/api/services/NetworkService.java
```diff
@@ -102,7 +105,7 @@ public JSONObject getLiveNIDS() {
 	public JSONObject getPeer(String id) {
 		try {
 			EUID euid = EUID.valueOf(id);
-			if (euid.equals(this.localPeer.euid())) {
+			if (euid.equals(EUID.fromHash(hasher.hash(this.localPeer)))) {
```
clstrfsck: Would be better to compute and store EUID for localPeer in constructor here, rather than recomputing & hashing for each request.

radixdlt/src/main/java/org/radix/network/messages/PeersMessage.java
```diff
@@ -15,48 +15,62 @@
...
+import com.radixdlt.serialization.SerializerId2;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@SerializerId2("network.message.peers")
+public final class PeersMessage extends Message
+{
+	@JsonProperty("peers")
+	@DsonOutput(Output.ALL)
+	private ImmutableList<Peer> peers = ImmutableList.of();
```
clstrfsck: I think there is a bug here.  In PeerManager.java, circa line 267, we have the code:
```
peersMessage.getPeers().add(p);
```

This change clearly breaks that quite badly.  Can we modify the code in PeersManager to construct with an ImmutableList once the list is built, and get rid of setPeers(...) altogether?

Something like this:
```
// Constructor
public PeersMessage(int magic, Collection<Peer> peers) {
    super(magic);
    this.peers = ImmutableList.copyOf(peers);
}
```

Code in PeersManager
```
// Deliver known Peers in its entirety, filtered on whitelist and activity
// Chunk the sending of Peers so that UDP can handle it
List<Peer> peersList = Lists.newArrayList();
List<Peer> peers = addressbook.peers()
	.filter(Peer::hasNID)
	.filter(StandardFilters.standardFilter(localSystem.getNID(), whitelist))
	.filter(StandardFilters.recentlyActive(this.recencyThreshold))
	.collect(Collectors.toList());

for (Peer p : peers) {
	if (p.getNID().equals(peer.getNID())) {
		// Know thyself
		continue;
	}

	peersList.add(p);
	if (peersList.size() == peerMessageBatchSize) {
		messageCentral.send(peer, new PeersMessage(this.universeMagic, peersList));
		peersList.clear();
	}
}

if (!peersList.isEmpty()) {
	messageCentral.send(peer, new PeersMessage(this.universeMagic, peersList));
}
```

radixdlt/src/main/java/org/radix/network/messaging/SignedMessage.java
```diff
@@ -49,16 +50,15 @@ public final void setSignature(ECDSASignature signature) {
 		this.signature = signature;
 	}
 
-
-	public void sign(ECKeyPair key) {
-		sign(key, false);
+	public static void sign(SignedMessage signedMessage, ECKeyPair key, Hasher hasher) {
```
clstrfsck: Not sure these are all static methods now?

radixdlt/src/main/java/org/radix/universe/system/SystemMessage.java
```diff
@@ -41,10 +43,27 @@ public SystemMessage(RadixSystem system, int magic) {
...
-	}
+	}
+
+	@Override
+	public String toString() {
+		return String.format("%s[%s]", getClass().getSimpleName(), this.system.getNID());
+	}
+
+	@Override
+	public boolean equals(Object o) {
```
clstrfsck, line 54: Needs EqualsVerifier test.

radixdlt/src/test/java/com/radixdlt/ledger/DtoCommandsAndProofTest.java
```diff
@@ -0,0 +1,15 @@
+package com.radixdlt.ledger;
```
clstrfsck, line 18: Missing header comment.

radixdlt/src/test/java/com/radixdlt/network/messaging/MessageDispatcherTest.java
```diff
@@ -63,12 +67,32 @@
...
 	}
 
 	static class DummyMessage extends Message {
 		DummyMessage() {
 			super(0);
 		}
+
+		@Override
+		public boolean equals(Object o) {
+			return o == null || getClass() != o.getClass();
```
clstrfsck: This seems wrong, but not sure it matters here.  Perhaps just using default identity hash and equals would be fine, and just removing these methods?

radixdlt/src/test/java/com/radixdlt/network/messaging/MessageDispatcherTest.java
```diff
@@ -63,12 +67,32 @@
 		DummySignedMessage() {
 			super(0);
 		}
+
+		@Override
+		public boolean equals(Object o) {
+			return o == null || getClass() != o.getClass();
```
clstrfsck: This seems wrong, but not sure it matters here.  Perhaps just using default identity hash and equals would be fine, and just removing these methods?

clstrfsck: Probably should be 
```
return o != null && getClass() == o.getClass();
```
if we want them all to be equal.

clstrfsck (changes requested):
It looks like this merge will break PeersManager as noted in comments, due to some less than optimal code there.  We should fix that, suggestion included.

Other minor comments.

radixdlt/src/main/java/org/radix/api/services/NetworkService.java
LukasGasior1: yes, definitely

radixdlt/src/main/java/org/radix/network/messages/PeersMessage.java
LukasGasior1: a similar change in PeerManager was already included in this PR: https://github.com/radixdlt/radixdlt-core/pull/230/files#diff-ff41b2829e41459497a0d7c04df7fe50R268

I'll change it to what you suggested though

radixdlt/src/main/java/org/radix/api/services/NetworkService.java
LukasGasior1: I've added another class var

radixdlt/src/main/java/com/radixdlt/CryptoModule.java
LukasGasior1: his change is already in the base branch (removed line above)

siy (approved):
LGTM

radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/application/RadixEngineValidatorRegistrator.java
```diff
@@ -76,8 +80,8 @@ public Command nextCommand() {
 		atom.addParticleGroup(particleGroup);
 		final Command command;
 		try {
-			atom.sign(keyPair);
-			ClientAtom clientAtom = ClientAtom.convertFromApiAtom(atom);
+			Atom.sign(atom, keyPair, hasher);
```
talekhinezh: Didn't we revert back to non-static methods?

radixdlt/src/main/java/com/radixdlt/middleware2/ClientAtom.java
```diff
@@ -90,15 +92,18 @@ private ClientAtom(
...
 
 	@JsonProperty("raw")
-	private void setSerializerAtom(byte[] atomBytes) {
-		Objects.requireNonNull(atomBytes);
+	private void setSerializerAtom(byte[] rawAtom) {
+		Objects.requireNonNull(rawAtom);
+		this.rawAtom = rawAtom;
+	}
+
+	public void init(Hasher hasher, Serialization serialization) {
```
talekhinezh: ClientAtom is an immutable POJO style object. Having a method which modifies it's state breaks it's immutability. Would rather keep the global hasher for now rather than introducing stateful POJOs.

radixdlt/src/main/java/com/radixdlt/consensus/Sha256Hasher.java
```diff
@@ -0,0 +1,48 @@
...
+package com.radixdlt.consensus;
+
+import com.google.common.hash.HashCode;
+import com.radixdlt.DefaultSerialization;
+import com.radixdlt.crypto.HashUtils;
+import com.radixdlt.crypto.Hasher;
+import com.radixdlt.serialization.DsonOutput;
+import com.radixdlt.serialization.Serialization;
+
+public class Sha256Hasher implements Hasher {
```
talekhinezh, line 30: Missing javadoc

radixdlt/src/main/java/com/radixdlt/consensus/bft/BFTBuilder.java
```diff
@@ -20,7 +20,7 @@
 import com.radixdlt.consensus.BFTEventProcessor;
 import com.radixdlt.consensus.HashSigner;
 import com.radixdlt.consensus.HashVerifier;
-import com.radixdlt.consensus.Hasher;
+import com.radixdlt.crypto.Hasher;
```
talekhinezh, line 23: See comment about keeping Hasher.java in core

radixdlt/src/main/java/com/radixdlt/network/addressbook/PeerWithTransport.java
```diff
@@ -100,5 +101,20 @@ public String toString() {
 		return String.format("%s[%s]", getClass().getSimpleName(), transportInfo);
 	}
 
-	// Note that we rely on equals(...) and hashCode() from BasicContainer here.
+	@Override
```
talekhinezh, line 104: I think replacing all of these equals is too big of a change and producing unintended bugs. For example, this equals() is missing `ban_reason` from Peer.java. Would much prefer to retain global hasher for now and not have such a big change across the codebase.

talekhinezh (changes requested):

radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/application/RadixEngineValidatorRegistrator.java
LukasGasior1: yep, forgot to update this repo

radixdlt/src/main/java/com/radixdlt/middleware2/ClientAtom.java
LukasGasior1: how about doing it this way then (introducing an intermediate class):
```
class RawClientAtom {
   byte[] rawAtom;

   public ClientAtom toClientAtom() {
     return new ClientAtom(...init from raw data here...);
  }
}
```

radixdlt/src/main/java/com/radixdlt/network/addressbook/PeerWithTransport.java
LukasGasior1: I've omitted ban reason on purpose, thought it makes sense to keep them equal if transportInfo is the same. We have EqualsVerifier tests for all those - just did a quick search, Peer classes are the only ones with ignored fields (ban reason & timestamps) - I'll include them too then.

@clstrfsck any comment on Josh's suggestion to revert all those overriden equals and go back to a global hasher?

radixdlt/src/main/java/com/radixdlt/middleware2/ClientAtom.java
talekhinezh: I actually think a bit of this will be resolved in my PR: https://github.com/radixdlt/radixdlt-core/pull/231 as the hash itself will get serialized.

I think cleanest way to do this is to merge that branch into this one.

radixdlt/src/main/java/com/radixdlt/network/addressbook/PeerWithTransport.java
talekhinezh: ah yes, you're right ban_reason is Output.PERSIST which I don't think gets used in the hash.

talekhinezh (approved):

clstrfsck (approved):

PR #231, opened 2020-10-12
Implement system update into Radix Engine

talekhinezh:
* Every committed round of consensus updates radix engine
* Refactor ClientAtom


siy (approved):
LGTM

radixdlt/src/main/java/com/radixdlt/middleware2/ClientAtom.java
```diff
@@ -56,57 +62,158 @@
...
-		} catch (DeserializeException | LedgerAtomConversionException e) {
-			throw new IllegalStateException("Failed to deserialize atomBytes");
+	private ImmutableList<byte[]> getSerializerInstructions() {
+		return serializedInstructions(this.instructions).collect(ImmutableList.toImmutableList());
+	}
+
+	private static Stream<byte[]> serializedInstructions(ImmutableList<CMMicroInstruction> instructions) {
+		return instructions.stream().flatMap(i -> {
+			if (i.getMicroOp() == CMMicroOp.PARTICLE_GROUP) {
+				return Stream.of(new byte[] {0});
```
clstrfsck: Would be better to create these byte arrays outside of the loop so that we don't have more garbage than necessary.

clstrfsck (approved):
Minor issue with serialisation that could be straightforwardly changed to reduce garbage production.

LukasGasior1 (approved):
LGTM

PR #233, opened 2020-10-12
Feature/epoch high view in re

talekhinezh:


radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
```diff
@@ -195,19 +198,47 @@ private void commitCommand(long version, Command command, VerifiedLedgerHeaderAn
...
+		final long currentEpoch = lastSystemParticle.getEpoch();
+		boolean epochChange = false;
+
 		final VerifiedLedgerHeaderAndProof headerAndProof = verifiedCommandsAndProof.getHeader();
 		long stateVersion = headerAndProof.getAccumulatorState().getStateVersion();
 		long firstVersion = stateVersion - verifiedCommandsAndProof.getCommands().size() + 1;
 		for (int i = 0; i < verifiedCommandsAndProof.getCommands().size(); i++) {
 			this.commitCommand(firstVersion + i, verifiedCommandsAndProof.getCommands().get(i), headerAndProof);
+
+			long nextEpoch = radixEngine.getComputedState(SystemParticle.class).getEpoch();
```
LukasGasior1: ~can this be called outside of the loop?~ nvm

radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
```diff
@@ -195,19 +198,47 @@ private void commitCommand(long version, Command command, VerifiedLedgerHeaderAn
...
+			if (nextEpoch > currentEpoch) {
+				// TODO: revert commands and log as proof of byzantine quorum rather than throw illegal state exception
+				if (i != verifiedCommandsAndProof.getCommands().size() - 1) {
+					throw new ByzantineQuorumException("change of epoch did not occur on last command");
```
LukasGasior1: what's the effect of this throw?

radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
```diff
@@ -195,19 +198,47 @@ private void commitCommand(long version, Command command, VerifiedLedgerHeaderAn
...
+				}
+				epochChange = true;
+				break;
```
LukasGasior1: is this break needed, since epoch change can only happen in the last command anyway?

btw, I'm a big fan of exhaustive if-else statements (and trying to avoid nesting, if possible), what do you think about this version?
```
if (isLastCommand && commandChangedEpoch) {
  epochChange = true;
} else if (commandChangedEpoch) {
  throw new ByzantineQuorumException();
} else {
  continue;
}
```

clstrfsck (approved):
Looks good.

radixdlt/build.gradle
```diff
@@ -139,7 +139,7 @@ configurations.all {
 }
 
 dependencies {
-    compile 'com.radixdlt:radix-engine-library:rc~1.0-beta.17-SNAPSHOT'
+    compile 'com.radixdlt:radix-engine-library:feature~epoch-high-view-SNAPSHOT'
```
siy: If changes for radix-engine-library will be merged first, this change is unnecessary. Also, it potentially may break other PR's.

radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
```diff
@@ -63,20 +64,20 @@
...
+	private final View epochCeilingView;
 
 	public RadixEngineStateComputer(
 		Serialization serialization,
 		RadixEngine<LedgerAtom> radixEngine,
-		View epochChangeView
+		View epochCeilingView
 	) {
-		if (epochChangeView.isGenesis()) {
+		if (epochCeilingView.isGenesis()) {
```
siy, line 90: Perhaps it better to use factory method and check preconditions there instead of throwing exception in constructor.

siy (changes requested):

radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
talekhinezh: At the moment, this just kills the system as we have no mechanism with dealing with byzantine quorums

radixdlt/build.gradle
talekhinezh: yes, will change this before merging

radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
talekhinezh: What's the reasoning for doing it in that way?

radixdlt/src/main/java/com/radixdlt/statecomputer/RadixEngineStateComputer.java
talekhinezh: I like it, thanks

LukasGasior1 (approved):

PR #238, opened 2020-10-21
Add memory leak detector

LukasGasior1:
Added a simple memory leak detector that keeps track of post-GC old gen heap, and alerts when it's steadily increasing.

### DIFF
--- a/radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
+++ b/radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
@@ -1,0 +1,87 @@
+package com.radixdlt.utils;
+
+import com.sun.management.GarbageCollectionNotificationInfo;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import javax.management.Notification;
+import javax.management.NotificationEmitter;
+import javax.management.NotificationListener;
+import javax.management.openmbean.CompositeData;
+import java.lang.management.GarbageCollectorMXBean;
+import java.lang.management.ManagementFactory;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class MemoryLeakDetector {
+
+    private static final Logger log = LogManager.getLogger();
+
+    private static final String OldGenGcName = "G1 Old Gen";
+
+    private static final int CONSECUTIVE_INCREASE_ALERT_THRESHOLD = 3;
+
+    private final List<Long> memTrail = new ArrayList<>();
+
+    public void init() {
+        log.info("Starting memory leak detector...");
+        final List<GarbageCollectorMXBean> garbageCollectorMXBeans = ManagementFactory.getGarbageCollectorMXBeans();
+        for (GarbageCollectorMXBean garbageCollectorMXBean : garbageCollectorMXBeans) {
+            listenForGarbageCollectionOn(garbageCollectorMXBean);
+        }
+    }
+
+    private void listenForGarbageCollectionOn(GarbageCollectorMXBean garbageCollectorMXBean) {
+        final NotificationEmitter notificationEmitter = (NotificationEmitter) garbageCollectorMXBean;
+        final GCListener listener = new GCListener();
+        notificationEmitter.addNotificationListener(listener, null, null);
+    }
+
+    private class GCListener implements NotificationListener {
+        @Override
+        public void handleNotification(Notification notification, Object handback) {
+            if (notification.getType().equals(GarbageCollectionNotificationInfo.GARBAGE_COLLECTION_NOTIFICATION)) {
+                final GarbageCollectionNotificationInfo info =
+                        GarbageCollectionNotificationInfo.from((CompositeData) notification.getUserData());
+
+                final long oldGenMemUsageBeforeGc = info.getGcInfo()
+                        .getMemoryUsageBeforeGc().get(OldGenGcName).getUsed();
+
+                final long oldGenMemUsageAfterGc = info.getGcInfo()
+                        .getMemoryUsageAfterGc().get(OldGenGcName).getUsed();
+
+                if (oldGenMemUsageAfterGc < oldGenMemUsageBeforeGc) { // we're only considering GC runs where old gen was cleaned
+                    memTrail.add(oldGenMemUsageAfterGc);
+                    if (memTrail.size() > CONSECUTIVE_INCREASE_ALERT_THRESHOLD + 1) {
+                        memTrail.remove(0); // to keep the list bounded
+                    }
+
+                    boolean consecutivelyIncreasing = true;
+                    for (int i = 1; i < memTrail.size(); i++) {
+                        if (memTrail.get(i) <= memTrail.get(i - 1)) {
+                            consecutivelyIncreasing = false;
+                            break;
+                        }
+                    }
+
+                    if (consecutivelyIncreasing && memTrail.size() > CONSECUTIVE_INCREASE_ALERT_THRESHOLD) {
+                        final String memTrailStr = memTrail.stream()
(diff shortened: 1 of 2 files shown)


radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
```diff
@@ -0,0 +1,87 @@
...
+                            consecutivelyIncreasing = false;
+                            break;
+                        }
+                    }
+
+                    if (consecutivelyIncreasing && memTrail.size() > CONSECUTIVE_INCREASE_ALERT_THRESHOLD) {
+                        final String memTrailStr = memTrail.stream()
+                                .map(n -> Long.toString(toMb(n)))
+                                .collect(Collectors.joining(", "));
+                        log.info("Potential memory leak detected! After {} latest GC runs the old gen heap steadily increases.",
```
LukasGasior1: up to discussion: what kind of alert do we want?

siy: I guess WARN should be fine here

siy (approved):
LGTM

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
```diff
@@ -0,0 +1,87 @@
+package com.radixdlt.utils;
```
clstrfsck, line 18: Missing header comment.

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
```diff
@@ -0,0 +1,87 @@
...
+import java.lang.management.ManagementFactory;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class MemoryLeakDetector {
+
+    private static final Logger log = LogManager.getLogger();
+
+    private static final String OldGenGcName = "G1 Old Gen";
```
clstrfsck: Should have constant name eg `OLD_GEN_GC_NAME`.

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private static final int CONSECUTIVE_INCREASE_ALERT_THRESHOLD = 3;
+
+    private final List<Long> memTrail = new ArrayList<>();
+
+    public void init() {
+        log.info("Starting memory leak detector...");
+        final List<GarbageCollectorMXBean> garbageCollectorMXBeans = ManagementFactory.getGarbageCollectorMXBeans();
+        for (GarbageCollectorMXBean garbageCollectorMXBean : garbageCollectorMXBeans) {
```
clstrfsck: Could be `garbageCollectorMXBeans.forEach(this::listenForGarbageCollectionOn)`, or even:
```
    ManagementFactory.getGarbageCollectorMXBeans()
        .forEach(this::listenForGarbageCollectionOn);
```

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
```diff
@@ -0,0 +1,87 @@
...
+
+public class MemoryLeakDetector {
+
+    private static final Logger log = LogManager.getLogger();
+
+    private static final String OldGenGcName = "G1 Old Gen";
+
+    private static final int CONSECUTIVE_INCREASE_ALERT_THRESHOLD = 3;
+
+    private final List<Long> memTrail = new ArrayList<>();
```
clstrfsck: Given that we are removing at the head, would be better to use a LinkedList here.

Another possibility is to use EvictingQueue from Guava, which works like a fixed-size circular buffer.

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
```diff
@@ -0,0 +1,87 @@
...
+                                toMb(oldGenMemUsageAfterGc), memTrailStr);
+
+                        memTrail.clear(); // reset the state
+                    }
+                }
+            }
+        }
+
+        private long toMb(long valueInBytes) {
+            return valueInBytes / 1000000;
```
clstrfsck: Could we use MiB here (2^20) rather than MB (10^6), as this is more likely to be consistent with other tools (eg visualvm).

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private static final Logger log = LogManager.getLogger();
+
+    private static final String OldGenGcName = "G1 Old Gen";
+
+    private static final int CONSECUTIVE_INCREASE_ALERT_THRESHOLD = 3;
+
+    private final List<Long> memTrail = new ArrayList<>();
+
+    public void init() {
```
clstrfsck: Not really a big fan of two-phase construction.  Can we move this into the constructor.

Also, we are missing a "stop" method that unhooks callbacks.  I think this is required if we want to integrate this tool into tests.

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
clstrfsck: WARN sounds good.

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    private void listenForGarbageCollectionOn(GarbageCollectorMXBean garbageCollectorMXBean) {
+        final NotificationEmitter notificationEmitter = (NotificationEmitter) garbageCollectorMXBean;
+        final GCListener listener = new GCListener();
+        notificationEmitter.addNotificationListener(listener, null, null);
+    }
+
+    private class GCListener implements NotificationListener {
```
clstrfsck: No real need to wrap this interface in a class.  It could just be a method, and then use a method reference in listenForGarbageCollectionOn(...).

Something like this:
```
    private void listenForGarbageCollectionOn(GarbageCollectorMXBean garbageCollectorMXBean) {
        final NotificationEmitter notificationEmitter = (NotificationEmitter) garbageCollectorMXBean;
        notificationEmitter.addNotificationListener(this::handleNotification, null, null);
    }

    public void handleNotification(Notification notification, Object handback) {
        // ...rest of code here
```

clstrfsck (changes requested):
Some minor comments, but would like to see these things fixed:

- Missing header comment
- Some way of stopping (and detaching listeners) so we can use this class in tests

radixdlt/src/main/java/com/radixdlt/utils/MemoryLeakDetector.java
LukasGasior1: :+1: I'll use LinkedList.
I've considered EvictingQueue but queues don't provide by-index access.

clstrfsck: Right, but it would be better to use an iterators-based ordering check rather than index-based.

Guava has `Comparators.isInStrictOrder(memTrail, Long::compare)` that gives the right answer and is iterator/iterable based.

LukasGasior1: Nice, I didn't know about that one. I've used it (together with EvictingQueue).

siy (approved):
LGTM

clstrfsck (approved):
Looks good.

PR #243, opened 2020-10-27
Additional particle validation for fee groups

LukasGasior1:


radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
```diff
@@ -129,26 +130,25 @@ private boolean isFeeGroup(ParticleGroup pg) {
 		List<SpunParticle> spunTransferableTokens = grouping.remove(TransferrableTokensParticle.class);
 		List<SpunParticle> spunUnallocatedTokens = grouping.remove(UnallocatedTokensParticle.class);
```
siy: Perhaps it would worth to use something like that:
`List<SpunParticle> spunTransferableTokens = Optional.ofNullable(grouping.remove(TransferrableTokensParticle.class)).orElseGet(List::of);`
This would simplify code downstream.

radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
```diff
@@ -129,26 +130,25 @@ private boolean isFeeGroup(ParticleGroup pg) {
...
-		if (grouping.isEmpty() && spunUnallocatedTokens != null) {
-			ImmutableList<TransferrableTokensParticle> transferableTokens = spunTransferableTokens == null
-				? ImmutableList.of()
-				: spunTransferableTokens.stream()
-					.map(SpunParticle::getParticle)
-					.map(p -> (TransferrableTokensParticle) p)
-					.collect(ImmutableList.toImmutableList());
-			return allUpForFeeToken(spunUnallocatedTokens) && allSameAddressAndForFee(transferableTokens);
+		if (grouping.isEmpty() && spunTransferableTokens != null && spunUnallocatedTokens != null) {
+			final ImmutableList<Pair<Spin, TransferrableTokensParticle>> transferableTokens =
```
siy: By inverting condition it's possible to reduce code nesting level.

radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
```diff
@@ -165,6 +165,26 @@ private boolean isUpAndForFee(SpunParticle sp) {
...
+		final Map<Spin, List<TransferrableTokensParticle>> particlesBySpin = transferableTokens.stream()
+				.collect(Collectors.groupingBy(Pair::getFirst, Collectors.mapping(Pair::getSecond, Collectors.toList())));
+
+		final List<TransferrableTokensParticle> inputParticles = particlesBySpin.getOrDefault(Spin.DOWN, List.of());
+		final List<TransferrableTokensParticle> outputParticles = particlesBySpin.getOrDefault(Spin.UP, List.of());
+
+		if (outputParticles.isEmpty()) {
+			return true;
+		} else {
+			final UInt256 outputAmount = outputParticles.get(0).getAmount();
```
siy: Code from `else` block can be moved out of block, so nesting will be reduced.

siy (changes requested):

radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
```diff
@@ -129,26 +130,25 @@ private boolean isFeeGroup(ParticleGroup pg) {
...
-			ImmutableList<TransferrableTokensParticle> transferableTokens = spunTransferableTokens == null
-				? ImmutableList.of()
-				: spunTransferableTokens.stream()
-					.map(SpunParticle::getParticle)
-					.map(p -> (TransferrableTokensParticle) p)
-					.collect(ImmutableList.toImmutableList());
-			return allUpForFeeToken(spunUnallocatedTokens) && allSameAddressAndForFee(transferableTokens);
+		if (grouping.isEmpty() && spunTransferableTokens != null && spunUnallocatedTokens != null) {
+			final ImmutableList<Pair<Spin, TransferrableTokensParticle>> transferableTokens =
+					spunTransferableTokens.stream()
```
clstrfsck: Not a big deal, but would prefer to see this as a single-tab indent.  IMO two-tab indents mean that systems (like github) that display tabs as 8 chars (by default) push content too far to the right to make it easily legible.

radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
LukasGasior1: heh, guess I need to adjust back to more imperative style of code ;) to me this:
```
if (abc) return cde
else return xyz;
```
looks much prettier than
```
if (abc) return cde;

return xyz;
```

radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
```diff
@@ -165,6 +165,26 @@ private boolean isUpAndForFee(SpunParticle sp) {
...
+
+		if (outputParticles.isEmpty()) {
+			return true;
+		} else {
+			final UInt256 outputAmount = outputParticles.get(0).getAmount();
+
+			final boolean allInputsGreaterThanOutput = inputParticles.stream()
+					.allMatch(ttp -> ttp.getAmount().compareTo(outputAmount) > 0);
+
+			return outputParticles.size() == 1 && allInputsGreaterThanOutput;
```
clstrfsck: Would prefer to see the constant-time check here performed before the linear-time check.

Something like:
```
	if (outputParticles.size() != 1) {
		return outputParticles.isEmpty();
	}
	final UInt256 outputAmount = outputParticles.get(0).getAmount();

	return inputParticles.stream()
		.allMatch(ttp -> ttp.getAmount().compareTo(outputAmount) > 0);
```

radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
```diff
@@ -165,6 +165,26 @@ private boolean isUpAndForFee(SpunParticle sp) {
 		return false;
 	}
 
+	// Check that there is at most one output particle and no input particle is smaller
+	private boolean noSuperfluousParticles(ImmutableList<Pair<Spin, TransferrableTokensParticle>> transferableTokens) {
+		final Map<Spin, List<TransferrableTokensParticle>> particlesBySpin = transferableTokens.stream()
+				.collect(Collectors.groupingBy(Pair::getFirst, Collectors.mapping(Pair::getSecond, Collectors.toList())));
```
clstrfsck: Could consider doing this grouping in isFeeGroup instead of creating a list of pairs, and then creating a map from that list.

allSameAddressAndForFee would need to be adjusted slightly, but overall would reduce number of collections created.

clstrfsck (approved):
Some minor feedback, but looks OK.

radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
LukasGasior1: good point

siy (approved):
LGTM

radixdlt/src/main/java/com/radixdlt/middleware2/TokenFeeLedgerAtomChecker.java
siy: It's not about functional vs imperative but rather "early cut off" approach. The idea is to make check which may prevent code from execution as soon as possible and leave function early, preserving "happy day scenario" at the top nesting level and therefore make it easier to read and follow.

PR #252, opened 2020-11-02
Fix for data race

clstrfsck:


radixdlt/src/main/java/org/radix/api/services/AtomsService.java
```diff
@@ -65,19 +68,21 @@
 import com.radixdlt.identifiers.AID;
 
 public class AtomsService {
-	private static int  NUMBER_OF_THREADS = 8;
+	private static final int  NUMBER_OF_THREADS = 8;
```
LukasGasior1: super minor: are there two spaces?

siy (approved):
LGTM

LukasGasior1 (approved):
LGTM

radixdlt/src/main/java/org/radix/api/services/AtomsService.java
clstrfsck: Yeap.  Fixed.

talekhinezh (approved):

PR #255, opened 2020-11-06
Task/rpnv1 262 release branch job

shambupujar:
These changes are related to below two JIRA tasks . As Release branches keeps changing, we need to setup multibranch job and add a jenkinsfile on the repo. 

This job https://jenkins-preprod.radixdlt.com/job/Release-Branch/job/build-core-multi-branch/ in setup to run for below branch and tag filters

release/1.0-beta* 1.0-beta*

Whenever a new branch/tag matching above is avaliable on github, then jenkin setups a job and run this jenkins file. There is bit of logic to check if the commit of the latest tag and the branch the job is running is same, (which happens when there is tag  added to release branch), then there is jenkins stage that pushes the image. Otherwise, it assumes the release branch has a new commit but has no tag, and just runs the jobs

https://radixdlt.atlassian.net/browse/RPNV1-690
https://radixdlt.atlassian.net/browse/RPNV1-262

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -87,16 +87,24 @@ pipeline {
         stage('Downstream jobs') {
             parallel {
                 stage('deploy-release'){
-                    build job: 'Release-Branch/deploy-release'
+                    steps{
+                        build job: 'Release-Branch/deploy-release'
+                    }
                 }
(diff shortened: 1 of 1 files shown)


siy (approved):
LGTM

clstrfsck (approved):
Looks good.

PR #261, opened 2020-11-12
Feature/deterministic sync

talekhinezh:


radixdlt/src/main/java/com/radixdlt/DispatcherModule.java
```diff
@@ -0,0 +1,116 @@
...
+	private EventDispatcher<BFTUpdate> viewEventDispatcher(
+		Environment environment
+	) {
+		return environment.getDispatcher(BFTUpdate.class);
+	}
+
+	@Provides
+	private EventDispatcher<EpochView> viewEventDispatcher(
+		@ProcessOnDispatch Set<EventProcessor<EpochView>> processors,
+		Set<EventProcessor<EpochView>> asyncProcessors,
```
clstrfsck, line 85: This argument seems to be ignored.  Is it required, or is this something that was removed at some point?

radixdlt/src/main/java/com/radixdlt/DispatcherModule.java
```diff
@@ -0,0 +1,116 @@
...
+				dispatcher.dispatch(epochView);
+				processors.forEach(e -> e.process(epochView));
+			};
+		}
+	}
+
+	@Provides
+	private EventDispatcher<Timeout> timeoutEventDispatcher(
+		@ProcessOnDispatch Set<EventProcessor<Timeout>> processors,
+		Set<EventProcessor<Timeout>> asyncProcessors,
```
clstrfsck, line 102: Similarly here.  This doesn't seem to be used, except for the emptiness checks.

radixdlt/src/main/java/org/radix/api/http/RadixHttpServer.java
```diff
@@ -23,7 +23,8 @@
 import com.radixdlt.crypto.Hasher;
 import com.radixdlt.consensus.bft.VerifiedVertex;
 import com.radixdlt.statecomputer.ClientAtomToBinaryConverter;
-import com.radixdlt.systeminfo.InMemorySystemInfoManager;
+import com.radixdlt.systeminfo.InMemorySystemInfo;
+import com.radixdlt.systeminfo.InMemorySystemInfoRunner;
```
clstrfsck: Not sure this import is used.

radixdlt/src/integration/java/com/radixdlt/integration/distributed/deterministic/DeterministicTest.java
```diff
@@ -206,6 +209,7 @@ public void configure() {
 			modules.add(new MockedCryptoModule());
 			modules.add(new LedgerLocalMempoolModule(10));
 			modules.add(new DeterministicMempoolModule());
+			modules.add(new DispatcherModule());
```
clstrfsck, line 208: I don't think the logger is used anymore, but can't add comment to that line.

radixdlt/src/test/java/com/radixdlt/consensus/epoch/EpochManagerTest.java
```diff
@@ -101,15 +102,14 @@
```
clstrfsck: I don't think validatorSet is used in this class.

clstrfsck (approved):
Not sure what the intent is with asyncProcessors, but the contents seem to be ignored in the providers where they are used.  Assuming this is resolved somehow, this looks OK to merge.

radixdlt/src/main/java/com/radixdlt/DispatcherModule.java
talekhinezh: These are required so that if there are no asyncProcessors then we skip dispatch the event into the environment. This is currently important because we've got deterministic tests which check for specific events in order and we dont want to include unused events in that queue.

LukasGasior1 (approved):
that's a lot of code ;) looks good

radixdlt/src/main/java/com/radixdlt/consensus/sync/BFTSync.java
```diff
@@ -363,8 +359,10 @@ public void processGetVerticesErrorResponse(GetVerticesErrorResponse response) {
 
 		log.debug("SYNC_VERTICES: Received GetVerticesErrorResponse: {} highQC: {}", response, vertexStore.highQC());
 
-		// error response indicates that the node has moved on from last sync so try and sync to a new sync
-		this.syncToQC(response.highQC(), response.getSender());
+		if (response.highQC().highestQC().getView().compareTo(vertexStore.highQC().highestQC().getView()) > 0) {
```
siy, line 362: It worth to extract condition check into dedicated method with proper name. This will simplify reading and eliminate the need for clarification comment.

siy (approved):
LGTM with one minor comment

PR #263, opened 2020-11-13
Fix DB configuration

siy:
Changed DB environment configuration according to suggestions from Dan. It is expected that this update will reduce amount of space consumed by DB on testnets.


clstrfsck (approved):

Theoklitos:
Looks good from what I can tell. Hopefully this gets merged today.

I will be looking at the metrics and let you know if this fixes the issue.

PR #267, opened 2020-11-18
Feature/rpnv1 742 validator selection

clstrfsck:
Setup and use validator registration and staking from genesis atom.


clstrfsck:
Needs integration branch

PR #269, opened 2020-11-18
[RPNV1-799] Dockerfile.test for test container

shambupujar:


siy (approved):
LGTM

docker/Dockerfile.test
```diff
@@ -0,0 +1,33 @@
+FROM ubuntu:16.04
```
clstrfsck: We need to stop using Ubuntu 16.04.  LTS support for this runs out in April next year, and that is not a time that we want to be scrambling to change things.

gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.jvmargs=-Xmx4g -XX:MaxPermSize=2048m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
clstrfsck: 4Gb seems completely over the top for a build system.  Do we know what this number really should be?

gradlew
```diff
@@ -44,7 +44,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS='"-Xmx64m" "-Xms64m"'
+DEFAULT_JVM_OPTS='"-Xmx2048m" "-Xms1024m"'
```
clstrfsck: Another 2GiB of RAM here means that we are using 6GiB of ram for the build system, if we account for a single daemon.

This seems completely ridiculous.  This is many, many times more RAM than the entire source code + build products (total of ~170MiB).

clstrfsck (changes requested):

docker/Dockerfile.test
```diff
@@ -0,0 +1,33 @@
...
+CMD /bin/bash
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends wget curl iputils-ping dnsutils \
+    unzip software-properties-common ca-certificates-java \
+    docker.io
+#install jdk11
+RUN add-apt-repository -y ppa:openjdk-r/ppa && \
+    apt update && \
+    apt install -yq openjdk-11-jdk
```
clstrfsck, line 18: Is there a headless package we could install?  Perhaps this would reduce the number of XWindows packages we need to pull in.

clstrfsck (approved):
Looks OK, but maybe installing the JDK headless package will reduce disk usage from unnecessary libraries for us?

PR #282, opened 2020-11-26
Mitigate OOM with rate limit on error responses

siy:


radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralValidatorSync.java
```diff
@@ -165,8 +174,17 @@ public void sendGetVerticesErrorResponse(BFTNode node, HighQC highQC) {
 					return null;
 				}
 
-				BFTNode node = BFTNode.create(src.getSystem().getKey());
-				return new GetVerticesErrorResponse(node, msg.highQC());
+				if (errorResponseRateLimiter.tryAcquire()) {
+					BFTNode node = BFTNode.create(src.getSystem().getKey());
+					return new GetVerticesErrorResponse(node, msg.highQC());
+				} else {
+					var counter = counters.add(CounterType.NETWORKING_DROPPED_ERROR_RESPONSES, 1);
```
clstrfsck: Should be increment(...) rather than add(..., 1).

radixdlt/src/main/java/com/radixdlt/counters/SystemCounters.java
```diff
@@ -70,6 +70,8 @@
 		NETWORKING_SENT_BYTES("networking.sent_bytes"),
 		NETWORKING_RECEIVED_BYTES("networking.received_bytes"),
 
+		NETWORKING_DROPPED_ERROR_RESPONSES("networking.dropped_error_responses"),
```
clstrfsck, line 73: Would be better as vertex_sync.dropped_error_responses, as networking.* is really the low-level stuff that talks IP.

Maybe @talekhinezh has some suggestions.

radixdlt/src/main/java/com/radixdlt/middleware2/network/ErrorRateLimit.java
```diff
@@ -0,0 +1,36 @@
...
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Identifies that the target represents a rate limit for incoming error messages.
+ */
+@Qualifier
+@Target({ FIELD, PARAMETER, METHOD })
+@Retention(RUNTIME)
+public @interface ErrorRateLimit {
```
clstrfsck: Would prefer this was a more specific name, perhaps GetVerticesErrorRateLimit

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralValidatorSync.java
```diff
@@ -165,8 +174,17 @@ public void sendGetVerticesErrorResponse(BFTNode node, HighQC highQC) {
...
+
+					if (counter % 1000 == 0) {
```
clstrfsck: Would be better as `== 1` so we don't have to wait for 1000 messages to be dropped before seeing the warning.

clstrfsck (approved):
The log criteria needs fixing so we don't have to wait for 1000 drops before we see a warning in the logs.  Other comments can be attended to as you see fit.

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralValidatorSync.java
siy: Fixed

radixdlt/src/main/java/com/radixdlt/middleware2/network/MessageCentralValidatorSync.java
siy: Added logging of first messages.

radixdlt/src/main/java/com/radixdlt/middleware2/network/ErrorRateLimit.java
siy: Renamed

radixdlt/src/main/java/com/radixdlt/counters/SystemCounters.java
siy: Well, it was hard to choose since this counter is updated at the point where we converting low level transport messages into internal objects. Messages are dropped before they are converted, so selection of the proper group for counter even more complicated.

talekhinezh: Yes, I think this is fine for now and practically speaking does what we need to do. I think we leave design work around this area when we gather more information.

talekhinezh (approved):

PR #291, opened 2020-12-08
Reduce frequency of unnecessary log

clstrfsck:
Reduce the frequency of these logs:
```
2020-12-08T09:12:12,062 [DEBUG/BFTEventPreprocessor/ConsensusRunner] (BFTEventPreprocessor.java:103) - ViewUpdate: Clearing Queues: {}
```

Which are produced at a rate of approximately 250,000/hour or 70/second.


siy (approved):
LGTM

PR #292, opened 2020-12-08
Feature/timeout certificates

LukasGasior1:
In progress


radixdlt/src/main/java/com/radixdlt/consensus/BFTHeader.java
```diff
@@ -47,6 +47,7 @@
 	@DsonOutput(Output.ALL)
 	private final HashCode vertexId;
 
+	// TODO(luk): rename this json property?
 	@JsonProperty("ledger_state")
```
talekhinezh: Yes go for it

radixdlt/src/main/java/com/radixdlt/consensus/HighQC.java
```diff
@@ -85,16 +97,34 @@ public static HighQC from(QuorumCertificate qc) {
...
+	public Optional<TimeoutCertificate> highestTC() {
+		return this.highestTC;
 	}
 
 	public QuorumCertificate highestQC() {
 		return this.highestQC;
 	}
 
+	public View getHighestView() {
+		if (this.highestTC.isPresent()
```
talekhinezh: Seems to be a prime place to implement this with monads :)

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -77,54 +80,105 @@ public boolean equals(Object obj) {
...
-			log.info("Ignoring vote from invalid author {}", node::getSimpleName);
-			return Optional.empty();
+			return VoteProcessingResult.rejected(VoteRejectedReason.INVALID_AUTHOR);
 		}
 
+		if (!replacePreviousVote(node, voteView, voteDataHash)) {
+			return VoteProcessingResult.rejected(VoteRejectedReason.DUPLICATE_VOTE);
+		}
+
+		if (!this.self.equals(nextProposer) && !vote.isTimeout()) {
```
talekhinezh: This should be checked outside of PendingVotes. Also remove reference to self. `PendingVotes` should not need to know who itself is nor next proposer logic.

radixdlt/src/main/java/com/radixdlt/consensus/bft/VertexStore.java
```diff
@@ -232,30 +237,49 @@ private VerifiedVertexStoreState getState() {
...
+
 	/**
 	 * Inserts a vertex and then attempts to create the next header.
 	 * If the ledger is ahead of the vertex store then returns an empty optional
 	 * otherwise an empty optional.
 	 *
 	 * @param vertex vertex to insert
 	 */
-	public void insertVertex(VerifiedVertex vertex) {
+	public Optional<BFTInsertUpdate> insertVertex(VerifiedVertex vertex) {
```
talekhinezh: Should not be exposing result of method here. Need to wait for async BFTInsertUpdate event. This is done for two reasons:
1. So we can implement async I/O down the line as this method requires a persistent store on success
2. So we can run recovery deterministic tests where failure of node can be simulated to be occurring with higher granularity between the message/processing async abstraction

talekhinezh (changes requested):

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
LukasGasior1: done

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -85,46 +86,90 @@ public PendingVotes(Hasher hasher) {
...
+		final Optional<QuorumCertificate> maybeQC = processVoteForQC(vote, validatorSet);
+		final Optional<TimeoutCertificate> maybeTC = processVoteForTC(vote, validatorSet);
+
+		if (maybeQC.isPresent()) {
+			return VoteProcessingResult.qcQuorum(maybeQC.get());
+		} else if (maybeTC.isPresent()) {
+			return VoteProcessingResult.tcQuorum(maybeTC.get());
+		} else {
+			return VoteProcessingResult.accepted();
+		}
```
siy:
```
		return maybeQC.<VoteProcessingResult>map(VoteProcessingResult::qcQuorum)
			.or(() -> maybeTC.map(VoteProcessingResult::tcQuorum))
			.orElseGet(VoteProcessingResult::accepted);
```

radixdlt/src/main/java/com/radixdlt/environment/RemoteEventDispatcher.java
```diff
@@ -19,11 +19,19 @@
...
+
 /**
  * Attempts to dispatch an event to a remote node.
  *
  * @param <T> the event class
  */
 public interface RemoteEventDispatcher<T> {
 	void dispatch(BFTNode receiver, T t);
+
+	default void dispatch(Set<BFTNode> receivers, T t) {
```
siy: Perhaps it worth to use `Iterable` instead of `Set`. All we need from this container is the ability to iterate over elements contained inside. Other `Set` properties like uniqueness of elements are irrelevant here.

siy (approved):
LGTM

radixdlt/src/main/java/com/radixdlt/environment/RemoteEventDispatcher.java
LukasGasior1: yeah, definitely

radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/application/TimestampChecker.java
```diff
@@ -32,7 +32,7 @@
  * to the real wall clock time.
  */
 public final class TimestampChecker implements TestInvariant {
-	private static final long ACCEPTABLE_TIME_RANGE = TimeUnit.SECONDS.toMillis(1);
+	private static final long ACCEPTABLE_TIME_RANGE = TimeUnit.SECONDS.toMillis(2);
```
clstrfsck: Better here would be to add an acceptable time range to a constructor, and make the test setup specify what this should be.

The we could add a `.addTimestampChecker(Duration maxDelay, String invariantName)` that could be tailored to suit the test in question, rather than making a global change here, which may invalidate some tests.  The existing `.addTimestampChecker(String invariantName)` could be changed to default to 1 sec.

In the case here, I think we would like to ensure that all commits happen within one timeout value, but for most tests, this intervening timeout doesn't occur.

radixdlt/src/integration/java/com/radixdlt/integration/distributed/simulation/tests/consensus_ledger_epochs/OneNodeNeverSendEpochResponseTest.java
```diff
@@ -64,7 +64,7 @@
 			for (long i = 0; i < epoch; i++) {
 				random.nextInt(numNodes);
 			}
-			return IntStream.range(0, random.nextInt(numNodes - 1) + 2)
+			return IntStream.range(0, random.nextInt(numNodes - 3) + 4)
```
clstrfsck: Actually, I'm not sure I like this that much.  It seems like there have been a few hacks/copy-pasta accumulated on top of each other that is making this much complex and harder to read than it needs to be.

1. I'm really not sure why there are two methods here (goodRandomEpochToNodesMapper and randomEpochToNodesMapper).  They should be folded into a single method.  Suspect this is a copy-pasta issue.
2. I'm unconvinced that the prefetching of randomness in the for loop is really doing anything useful, given that we are seeding the random with the epoch.
3. The method of selecting a random set of node ids seems overcomplicated.
4. We should be using manifest constants, rather than numeric literals for values.

IMO this code should look something like this:
```
private static final int minValidators = 4; // Need at least f=1 for this test

// ... code

private static Function<Long, IntStream> goodRandomEpochToNodesMapper() {
	return epoch -> {
		final var indices = IntStream.range(0, numNodes).boxed().collect(Collectors.toList());
		final var random = new Random(epoch);
		Collections.shuffle(indices, random);
		final var numValidators = minValidators + random.nextInt(numNodes - minValidators + 1);
		return indices.subList(0, numValidators).stream().mapToInt(Integer::intValue);
	};
}
```

radixdlt/src/main/java/com/radixdlt/consensus/HighQC.java
```diff
@@ -47,15 +49,22 @@
 	@DsonOutput(Output.ALL)
 	private final QuorumCertificate highestCommittedQC;
 
+	private final Optional<TimeoutCertificate> highestTC;
```
clstrfsck: Would prefer to see this stored as a nullable TimeoutCertificate, as use of this is entirely internal to the class.

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -85,46 +86,86 @@ public PendingVotes(Hasher hasher) {
...
-			return Optional.empty();
+			return VoteProcessingResult.rejected(VoteRejectedReason.INVALID_AUTHOR);
 		}
 
+		if (!replacePreviousVote(node, voteView, voteDataHash)) {
+			return VoteProcessingResult.rejected(VoteRejectedReason.DUPLICATE_VOTE);
+		}
+
+		final Optional<QuorumCertificate> maybeQC = processVoteForQC(vote, validatorSet);
+		final Optional<TimeoutCertificate> maybeTC = processVoteForTC(vote, validatorSet);
```
clstrfsck: Shouldn't construction of maybeTC (which may include expensive hashing, etc) be deferred until after we know that maybeQC is empty?

```
return processVoteForQC(vote, validatorSet)
	.<VoteProcessingResult>map(VoteProcessingResult::qcQuorum)
	.or(() -> processVoteForTC(vote, validatorSet).map(VoteProcessingResult::tcQuorum))
	.orElseGet(VoteProcessingResult::accepted);
```

radixdlt/src/main/java/com/radixdlt/consensus/Proposal.java
```diff
@@ -55,21 +56,33 @@
 	@DsonOutput(Output.ALL)
 	private final QuorumCertificate committedQC;
 
+	private final Optional<TimeoutCertificate> highestTC;
```
clstrfsck: I think it would be better if the internal representation here was a nullable TimeoutCertificate.

radixdlt/src/main/java/com/radixdlt/consensus/bft/BFTEventReducer.java
```diff
@@ -135,23 +158,41 @@ public void processBFTRebuildUpdate(BFTRebuildUpdate update) {
...
+		} else if (result instanceof VoteProcessingResult.VoteRejected) {
+			log.trace("Vote has been rejected because of: {}",
+				((VoteProcessingResult.VoteRejected) result).getReason());
+		} else if (result instanceof VoteProcessingResult.QuorumReached) {
+			this.hasReachedQuorum = true;
+			final ViewVotingResult viewResult =
+				((VoteProcessingResult.QuorumReached) result).getViewVotingResult();
+			viewQuorumReachedEventDispatcher
+				.dispatch(new ViewQuorumReached(viewResult, vote.getAuthor()));
+		}
```
clstrfsck, line 198: Really should have a exceptional failure case here if result is none of the known types.

radixdlt/src/main/java/com/radixdlt/consensus/bft/VertexStore.java
```diff
@@ -79,6 +82,7 @@ private VertexStore(
 		this.highestQC = Objects.requireNonNull(highestQC);
 		this.highestCommittedQC = Objects.requireNonNull(commitQC);
 		this.vertexChildren.put(rootVertex.getId(), new HashSet<>());
+		this.highestTC = highestTC;
```
clstrfsck: Should be a null check here `this.highestTC = Objects.requireNonNull(highestTC)`

radixdlt/src/main/java/com/radixdlt/consensus/bft/VoteProcessingResult.java
```diff
@@ -0,0 +1,148 @@
...
+
+import java.util.Objects;
+
+/**
+ * The result of processing a received vote.
+ */
+public interface VoteProcessingResult {
+
+    static VoteAccepted accepted() {
+        return new VoteAccepted();
```
clstrfsck: Can we use a singleton here, rather than creating a fresh object for each accepted vote?

radixdlt/src/main/java/com/radixdlt/consensus/safety/SafetyState.java
```diff
@@ -17,55 +17,87 @@
...
 /**
  * The state maintained to ensure the safety of the consensus system.
  */
+@Immutable
+@SerializerId2("consensus.safety_state")
 public final class SafetyState {
-	private final View lastVotedView; // the last view this node voted on (and is thus safe)
+
+	@JsonProperty(SerializerConstants.SERIALIZER_NAME)
+	@DsonOutput(value = {DsonOutput.Output.API, DsonOutput.Output.WIRE, DsonOutput.Output.PERSIST})
```
clstrfsck: This should be `@ DsonOutput(Output.ALL)`

radixdlt/src/main/java/com/radixdlt/environment/RemoteEventDispatcher.java
```diff
@@ -26,4 +26,10 @@
  */
 public interface RemoteEventDispatcher<T> {
 	void dispatch(BFTNode receiver, T t);
+
+	default void dispatch(Iterable<BFTNode> receivers, T t) {
+		for (BFTNode receiver: receivers) {
```
clstrfsck: Maybe `receivers.forEach(r -> dispatch(r, t))`

clstrfsck (approved):

radixdlt/src/main/java/com/radixdlt/consensus/PendingVotes.java
```diff
@@ -85,46 +86,86 @@ public PendingVotes(Hasher hasher) {
...
+			return Optional.of(new QuorumCertificate(voteData, validationState.signatures()));
+		} else {
 			return Optional.empty();
 		}
+	}
 
-		// If there is no equivocation or duplication, we process the vote.
-		ValidationState validationState = this.voteState.computeIfAbsent(voteDataHash, k -> validatorSet.newValidationState());
+	private Optional<TimeoutCertificate> processVoteForTC(Vote vote, BFTValidatorSet validatorSet) {
+		if (vote.getTimeoutSignature().isEmpty()) {
```
LukasGasior1: use vote.isTimeout

radixdlt/src/main/java/com/radixdlt/consensus/bft/BFTEventReducer.java
```diff
@@ -57,10 +57,18 @@
 	private BFTInsertUpdate latestInsertUpdate;
 	private ViewUpdate latestViewUpdate;
 
+	/* Indicates whether the quorum (QC or TC) has already been formed for the current view.
+	 * If the quorum has been reached (but view hasn't yet been updated), subsequent votes are ignored.
```
LukasGasior1, line 61: add comment

radixdlt/src/main/java/com/radixdlt/consensus/HighQC.java
LukasGasior1: okay, will change, but keep in mind that each time we use nulls a kitten dies - this one's on you :cat2:

radixdlt/src/main/java/com/radixdlt/consensus/Proposal.java
LukasGasior1: here goes another one :cat2:

radixdlt/src/main/java/com/radixdlt/consensus/liveness/Pacemaker.java
```diff
@@ -167,65 +184,96 @@ private void startView() {
...
+		if (this.timeoutVoteVertexId.isPresent()) {
+			return; // vertex for a timeout vote for this view is already inserted
+		}
+
+		final HighQC highQC = this.vertexStore.highQC();
+		final UnverifiedVertex proposedVertex = UnverifiedVertex.createVertex(highQC.highestQC(), view, null);
+		final VerifiedVertex verifiedVertex = new VerifiedVertex(proposedVertex, hasher.hash(proposedVertex));
+		this.timeoutVoteVertexId = Optional.of(verifiedVertex.getId());
+
+		this.vertexStore.getPreparedVertex(verifiedVertex.getId()).ifPresentOrElse(
```
talekhinezh, line 240: Can we create this empty vertex on new view creation like we talked about

talekhinezh (changes requested):

siy (approved):
LGTM

PR #312, opened 2020-12-22
Two small changes

Theoklitos:
Changed `java.rmi.server.hostname` to `core` so that CPU & memory monitoring gets fixed.

Added the iproute2 package, which will install [traffic control](https://manpages.debian.org/unstable/iproute2/tc.8.en.html), needed for some tests.


siy (approved):
LGTM

shambupujar (approved):
LGTM

PR #314, opened 2021-01-04
Sort nodes in RecoveryLivenessTest

LukasGasior1:


siy (approved):
LGTM

radixdlt/src/integration/java/com/radixdlt/integration/recovery/RecoveryLivenessTest.java
```diff
@@ -106,7 +107,10 @@
 	private NodeEvents nodeEvents;
 
 	public RecoveryLivenessTest(int numNodes, long epochCeilingView) {
-		this.nodeKeys = Stream.generate(ECKeyPair::generateNew).limit(numNodes).collect(Collectors.toList());
+		this.nodeKeys = Stream.generate(ECKeyPair::generateNew)
+			.limit(numNodes)
+			.sorted(Comparator.<ECKeyPair, EUID>comparing(k -> k.getPublicKey().euid()).reversed())
```
clstrfsck: Why is this reversed?

clstrfsck (commented):
Not sure why natural order of node IDs is reversed when sorting?

radixdlt/src/integration/java/com/radixdlt/integration/recovery/RecoveryLivenessTest.java
LukasGasior1: I did it for consistency with what we have in `DeterministicTest` [L132]; it may make debugging slightly easier if all tests use the same ordering. Not sure why it was originally reversed there though, it shouldn't matter if it's reversed or  not.

clstrfsck: Hmm.  Not sure it makes sense in DeterministicTest either, although git tells me I put that code in there ☹️ 

If it doesn't matter if it is reversed or not, why does the ordering matter at all then?

LukasGasior1: @clstrfsck We need an ordering that is deterministic in relation to proposer election (`ConsensusModule` sorts nodes by euid when creating `WeightedRotatingLeaders`). It doesn't matter if it's the same or reversed. The test will be deterministic across runs as long as there is any fixed relationship between these two orderings.

clstrfsck: Ah, OK.  Then is it worth simplifying this and DeterministicTest by removing the .reversed() ?

LukasGasior1: Probably yes, unless there's any hidden dependency on reversed order in DeterministicTest. I'll give it a try.

clstrfsck (approved):
Looks good, pending Travis approval.

